PR #30, opened 2018-08-02
Improve form field panel, Fixes #13, Fixes #24

### LINKED ISSUE #13: Add distinct examples of all three dropdown fields
We have three different dropdown fields which are core components and also different in important ways. Need examples of them in our Form tab.  We should also have examples of at ComboField with `requireSelection` true and false as that's another key setting to demonstrate and highlight.

Would suggest @TomTirapani and @febbraiod in terms of people on the team who could answer questions re. the fields or their behaviors if unclear.

### LINKED ISSUE #24: Consistency in Form Fields Tab
+ Need to display the value of *all* fields in the panel on the right.
+ Need to change confusion inducing "Profile Completion" field.  (its not actually linked)

febbraiod:


amcclain:
Thanks for this - I reworked the form fields pretty significantly around the same time as part of our larger updates, but I definitely grabbed updates / ideas from here as I did.  Closing as we've moved the files around.

PR #59, opened 2018-09-05
Add sub tab to demo grid with grouped columns 

febbraiod:


febbraiod:
Need to come up with a new (sane) data set to demo this. Rather than faking it in the trade service

febbraiod:
Lee and I agree we need to come up with a more reasonable dataset for a grouped col demo. Also this will depend on pending fixes to hoist react regarding colChooser and export of grouped col grids.

PR #60, opened 2018-09-05
Multi select field example

febbraiod:


febbraiod:
Dependent on associated Hoist-React PR

PR #62, opened 2018-09-05
(WIP) News feed

shravnk:
Toolbox news feed component, rendering using DataView
screen shot 2018-09-05 at 6 55 01 pm


amcclain:
Looks great - excellent progress. I am looking forward to seeing where it lands.

client-app/src/core/svc/NewsService.js
```diff
@@ -0,0 +1,26 @@
+/*
+ * This file belongs to Hoist, an application development toolkit
+ * developed by Extremely Heavy Industries (www.xh.io | info@xh.io)
+ *
+ * Copyright © 2018 Extremely Heavy Industries Inc.
+ */
+import {XH, HoistService} from '@xh/hoist/core';
+
+/**
```
lbwexler: Copy/Paste error in comment here .  Also, @HoistService() should be @HoistService

client-app/src/desktop/App.js
```diff
@@ -115,6 +117,14 @@ class AppClass {
                             {name: 'timestamp', path: '/timestamp'},
```
lbwexler: change 'features' to 'examples'

client-app/src/desktop/tabs/features/NewsPanel.js
```diff
@@ -0,0 +1,106 @@
...
+                tbar: toolbar({
+                    items: [
+                        button({
+                            text: 'Refresh Sources',
+                            icon: Icon.refresh(),
+                            onClick: this.onRefreshClick
+                        }),
+                        filler(),
+                        span({
+                            item: "Last updated:"
```
lbwexler: Favor Single quotes over double quotes

client-app/src/desktop/tabs/features/NewsPanelItem.js
```diff
@@ -0,0 +1,40 @@
...
+        const request = this.props.record,
+            {title, url, text, imageUrl, published, source, author} = request;
+
+        return hbox(
+            box({
+                flex: 4,
+                className: "text-container",
+                item: [
+                        <div>
+                            <p className="news-item--text">{source} | {published}</p>
```
lbwexler: Use single quotes

grails-app/services/io/xh/toolbox/App/NewsService.groovy
```diff
@@ -0,0 +1,88 @@
...
+    static clearCachesConfigs = ['newsSources', 'newsApiKey']
+    def configService
+
+    void init() {
+        log.info('news service initialized')
+        createTimer(
+                runFn: this.&loadAllNews,
+                interval: 10 * MINUTES
+        )
+        super.init()
```
lbwexler: The framework should already be logging this initialization.

If we were going to log this, we would log at *end* of method, and capitalize as, say. 'NewsService Initialized'.

client-app/src/desktop/tabs/features/NewsPanelModel.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    constructor() {
+        this.addReaction({
+            track: () => [this.sourceFilter, this.textFilter, this.lastRefresh],
+            run: () => this.filterData()
+        });
+    }
+
+    @action
```
lbwexler: Let's put filterData(),  completeLoad(), and destroy() below an implementation block marker

client-app/src/desktop/tabs/features/NewsPanel.js
```diff
@@ -0,0 +1,106 @@
...
+                            unit: 'stories'
+                        })
+                    ]
+                })
+            })
+        })
+    }
+
+    componentDidMount() {
+        this.model.loadData();
```
lbwexler: define loadAsync() here instead.  This is the root of a Tab.  This will also allow that the tab to play well with global refreshing of its parent container.

grails-app/init/io/xh/toolbox/BootStrap.groovy
```diff
@@ -49,6 +51,28 @@ class BootStrap {
             ]).save()
         }
     }
+    private void ensureRequiredConfigsCreated() {
+        Utils.configService.ensureRequiredConfigsCreated([
+                newsApiKey : [
+                        valueType   : 'string',
+                        defaultValue: '917b84920ec24d7198fd712e2e22b48a',
+                        groupName   : 'news',
+                        note        : 'Key used to access News API'
```
amcclain: Sorry to nitpick, but this is the kind of note/comment we can skip - it contains no new info relative to the name of the config itself.  It's OK not to have notes for these things if they are self-evident.

grails-app/services/io/xh/toolbox/App/NewsService.groovy
```diff
@@ -0,0 +1,88 @@
...
+
+        loadAllNews()
+    }
+
+
+    //------------------------
+    // Implementation
+    //------------------------
+    private void loadAllNews() {
+        def sources = configService.getJSONObject('newsSources', new JSONObject())
```
amcclain: We can remove the default value fallback here.  We want to use those very sparingly, and in the special case of Toolbox we know we will have it installed in Bootstrap.  In general, though, we prefer these things to throw and be obvious about it if they don't have the right configs setup.

(I realize this default was likely there in the old PIQ code.)

grails-app/services/io/xh/toolbox/BaseService.groovy
```diff
@@ -0,0 +1,54 @@
+package io.xh.toolbox
+
+import groovy.util.logging.Slf4j
+
+@Slf4j
+abstract class BaseService extends io.xh.hoist.BaseService {
```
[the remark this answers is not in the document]
amcclain: Definitely - this complexity is all related to working in a dataset / some code in PIQ that had a long history in another project. We can favor "just in time" additions of things like this - only add it when we need it.

grails-app/services/io/xh/toolbox/App/NewsService.groovy
```diff
@@ -0,0 +1,88 @@
...
+    public List<NewsItem> _newsItems
+
+    static clearCachesConfigs = ['newsSources', 'newsApiKey']
+    def configService
+
+    void init() {
+        log.info('news service initialized')
+        createTimer(
+                runFn: this.&loadAllNews,
+                interval: 10 * MINUTES
```
amcclain: This would be a nice place to use a feature of the `Timer` class where we can pass a string for the interval and have it read from a soft-config.  Our standard would define a config like `newsRefreshMins` where the config is an int representing the frequency in minutes.  That name (string) can then be passed here as the interval, along with an `intervalUnits: MINUTES` param to specify that the config is in minutes vs ms.

Although just tried to run locally and saw that we were rate limited out! I guess if PIQ is deployed and running it might be stealing hits from us - let's see if we can stop that....

PR #64, opened 2018-09-07
Use selectOnFocus in toolbox examples

nadkarniarjun:


lbwexler:
This PR has an incorrect name, but it actually seems reasonable, so I am going to merge

PR #65, opened 2018-09-10
Dedicated subtab to demo grouped columns in grids

febbraiod:


client-app/src/desktop/common/SampleGrid.js
```diff
@@ -126,7 +125,7 @@ class SampleGrid extends Component {
                     }),
                     storeCountLabel({
                         store,
-                        units: 'companies'
+                        unit: 'companies'
```
febbraiod: Not related to this project but figured I'd fix while I was here

PR #76, opened 2018-09-20
Demo RadioInput in form control panel

febbraiod:


client-app/src/desktop/tabs/forms/ControlsPanel.js
```diff
@@ -205,6 +206,17 @@ export class ControlsPanel extends Component {
...
+                                row({
+                                    label: 'RadioInput',
+                                    field: 'option6',
+                                    info: 'No soup for you',
+                                    item: radioInput({
+                                        label: 'Lunch',
+                                        align: 'right',
+                                        inline: true,
+                                        options: [{label: 'Soup', value: 'Soup', disabled: true}, 'Salad', 'Sandwich']
+                                    })
```
febbraiod: Don't necessarily think this is the demo of this control we'll want to take in the long run but wanted make it easy to see at least one of the quirks mentioned in the Hoist PR. Specifically the need for the style override in 'inline' mode.

PR #82, opened 2018-09-25
Add tree grid with checkboxes sample

cnrudd:


client-app/src/desktop/common/SampleTreeWithCheckBoxGrid.js
```diff
@@ -0,0 +1,228 @@
...
+            indeterminate = !allEnabled && !allDisabled;
+
+        rec.parent.indeterminate = indeterminate;
+        rec.parent.enabled = allEnabled;
+
+        this.updateParents(rec.parent);
+    }
+
+    sampleData() {
+        return [
```
lbwexler: Can we put this data in a shared file for sharing between the grids?  Maybe we will enhance it and make it more realistic.

cnrudd: sure, will do.

cnrudd: This is done.

PR #87, opened 2018-10-02
Update for dateIs could take special token for "now"

nadkarniarjun:
- Field validator dateIs could take special token for "now"

Updated Toolbox to show above as an example


client-app/src/desktop/tabs/forms/ValidationPanelModel.js
```diff
@@ -34,10 +34,10 @@ export class ValidationPanelModel {
     @field(numberIs({min: 0, max: 99}))
     yearsExperience;
 
-    @field('Hire Date', required, dateIs({min: moment().startOf('day').toDate()}))
+    @field('Hire Date', required, dateIs({max: 'now'}))
```
lbwexler: Seems like we will get into trouble with the time component here that is going to be included in 'now'.   I wonder if we need to add an additional token -- 'today' that would not have the time component.  

Note that our DateInput component now has a precision prop, that allows it to produce date only date times.   Looks like what we have on line 40 would not work the way we expect (e.g. control produces midnight on today, which will unexpectedly fail when compared to 'now'.

lbwexler:
Arjun -- could you take a look at my comment when you get a chance?  Would like to resolve and get this one in....

client-app/src/desktop/tabs/forms/ValidationPanelModel.js
nadkarniarjun: Done!

PR #93, opened 2018-10-15
Mobile agGrid

TomTirapani:
Use agGrid to power mobile grid.


client-app/src/mobile/grids/GridPageModel.js
```diff
@@ -19,23 +19,38 @@ export class GridPageModel {
...
-            headerName: 'P&L',
-            field: 'profit_loss',
-            valueFormatter: numberRenderer({
-                precision: 0,
-                ledger: true,
-                colorSpec: true,
-                asElement: true
-            })
-        }
+        subFields: [
```
lbwexler: where do I specify if I actually want the label for the SubField or not?  That was an important feature in zoned grid.  Also, whatever label we do have, can probably default to the field name, in StartCase()

PR #101, opened 2018-10-26
Add uses of labelFor and clickableLabel.

cnrudd:
Small changes to test hoist-react pr: [#753](https://github.com/exhi/hoist-react/pull/753)


lbwexler:
Hi Colin -- I am going to whack this, now that we removed the need for labelFor, and clickableLabel is defaulting to true.   (Don't know that we need to have examples of ClickableLabel false on this page).

Thanks for this fix.

PR #131, opened 2018-12-14
Added example of Select Input in Popover.

cnrudd:
Goes with PR https://github.com/exhi/hoist-react/pull/866


lbwexler:
This is more of a test of a bug-fix then a demo.  Removing as it is out-of-date, and looking to clean up our backlog here

Issue #128, opened 2018-12-13
Grid example with ag-Grid column filters


amcclain:
While proper support for ag-Grid provided column filters is still a pending item, we have enough real-world usages of these filters via the AG API that we need to be able to test and experiment with them in Toolbox.

We should add this as a dedicated grid in its own tab - i.e. specifically *don't* want to add them to the standard grid that's reused across examples as we can't fully recommend their use at this point.

shravnk:
See #139

lbwexler:
Closing this.  First class Hoist column filters will be included in v42.

PR #139, opened 2018-12-28
Sample column filter grid

shravnk:
For #128. Couple things to discuss:

1. I had used ag-grid's `onFilterChanged` to recalculate the number of grid records in another app. When testing it here, accessing the API that way caused each `ColumnHeader` to unmount and re-render, thus losing state and the correct filter icon. Discovered that the regression is due to Hoist 17.0/ag-Grid 19 update, but unclear why.

2. Think this would look a lot better if the header's menu icon were only displayed on hover. Think ag-grid supports this with a `Ref`, or maybe try custom CSS.

Both of these worthwhile I think, just trying to get a feel of how much we want to support these column filters


shravnk:
Updated with changes for 1) Required a one line bug fix which I pushed to Hoist develop.

Regarding 2), the I tested the below CSS. Could use it just for this example or make available in Hoist as well. 

`      .xh-grid-header-menu-icon {
        cursor: pointer;
        svg {
          visibility: hidden;
        }
        &:hover svg {
          visibility: visible;
        }
      }
`

lbwexler:
Brendan -- Can we please update this for review/merge?  There is still a little unrelated test 'readonly'  bit on restGrid that needs to come out, and also some copy/paste errors with the description.

shravnk:
Updated with Hoist 19 model pattern

PR #161, opened 2019-03-04
Make NewsService thread-safe using Hoist Cache

shravnk:


lbwexler:
Think this can be done more simply by just sorting the list before installing it, or setting the mutable false flag where we need to get a sorted copy of it.  (these sorts are mutable, and guaranteed to trip this non-threadsafe object up)

PR #167, opened 2019-04-04
Bind TreeGrid dimensions to a url parameter in TreeGrid example

TomTirapani:


client-app/src/desktop/common/SampleTreeGrid.js
```diff
@@ -134,5 +147,26 @@ class Model {
             .getPortfolioAsync(dims)
             .then(data => gridModel.loadData(data));
     }
+
+    syncDimsToRouter() {
+        const {name, params} = XH.routerState;
+        if (!name.endsWith('.tree')) return;
+
```
lbwexler: use router.isActive here?

client-app/src/desktop/common/SampleTreeGrid.js
```diff
@@ -134,5 +147,26 @@ class Model {
...
+            this.dimChooserModel.setValue(dims.split('.'));
+        }
+    }
+
+    syncRouterToDims(opts) {
+        const {name} = XH.routerState,
+            {dimChooserModel} = this,
+            dims = dimChooserModel.value.join('.');
+
+        if (!name.endsWith('.tree')) return;
```
lbwexler: use router.isActive Here?

lbwexler:
Thanks for this example Tom -- I was coincidentally up late last night futzing with TabContainer, and looking to dry up our global app-wide route definition into the TabContainerModel itself.

I got quite close, but think that's its ultimately futile -- we simply need the routes before any components/model are defined, because on page load, the routes may determine which components ultimately get loaded. 

 I had been toying with ways in which we might still actually create the routes in the Tab files themselves, but I think your example shows another reason why we generally want to have the routes defined at the app level.

PR #179, opened 2019-05-02
Recall Example App -- First Version!

YanaNirvana:


lbwexler:
Hi Yana -- as we went over this is a great little applet, and eager to get it in.

Two quick things, that we really should add before merging --

1)  Looks like we need to actually go back to reading the new config for that url!
2) Can we get some minimal text somewhere to describe the source?  Maybe a  little "about" button that pops up a little modal dialog.  See XH.message()

PR #182, opened 2019-05-09
New Recalls Panel with groupby and scrolling

YanaNirvana:
+ added buttonGroupInput to allow drug recall records to be grouped by fields such as: recalling firm, status, classification;
+ buttonGroupInput uses the new `enableClear: true` config so that users may unselect buttons
+ added scrolling to details panel
+ details panel now displays drug name if a drug is selected
+ co dev w Lee to sync up state between statemodel and gridmodel

Screen Shot 2019-05-09 at 11 30 00 AM


client-app/src/desktop/tabs/examples/recalls/RecallsPanelModel.js
```diff
@@ -6,16 +6,16 @@
...
-import moment from 'moment';
 import {dateCol} from '@xh/hoist/cmp/grid/columns';
 import {compactDateRenderer} from '@xh/hoist/format';
-import {managed} from '@xh/hoist/core/mixins';
 import {Icon} from '@xh/hoist/icon/Icon';
 import {ONE_SECOND} from '@xh/hoist/utils/datetime';
 
 import {DetailsPanelModel} from './DetailsPanelModel';
 import {bindable} from '@xh/hoist/mobx';
+import moment from 'moment';
```
lbwexler: is this import needed?

client-app/src/desktop/tabs/other/PopupsPanel.js
```diff
@@ -0,0 +1,21 @@
+/*
```
lbwexler: Woops -- looks like this file snuck in to this branch!

client-app/src/desktop/tabs/examples/recalls/RecallsPanelModel.js
YanaNirvana: Yupp-- we use moment in processRecord (new line 140).

Sorry for the confusion -- it looks like a new line when all I did was move it (see old line 10)

Issue #170, opened 2019-04-09
Grids > Group Columns: Columns stay "grouped" after page "refresh" function 


mpeckerman:
1. Navigate to Hoist Reach Toolbox -> Grids tab -> Group Column panel / screen
2. Turn "Group Rows" toggle switch ON
3. Refresh the screen
4. Rows in the grid are GROUPED, while the control switch is OFF

PR #184, opened 2019-05-10
Synced up groupBy between stateModel and model's toggle button

YanaNirvana:
#170


client-app/src/desktop/common/SampleColumnGroupsGrid.js
```diff
@@ -32,6 +32,7 @@ class SampleColumnGroupsGrid extends Component {
 
         return panel({
             item: grid({model: gridModel}),
+            ref: (elem) => this.model.panelRef = elem,
```
lbwexler: We have a Ref class that handles this nicely

Issue #174, opened 2019-04-18
Add Other › Popups examples tab


amcclain:
We have this in the mobile Toolbox, but would be useful to have on the desktop as well. A new tab under Other with buttons to launch our various popup alerts.  

Should have examples of `XH.alert()`, `XH.confirm()`, and `XH.toast()`, and include an example that makes it clear (if you look at the code) how to use the promise returned by XH.confirm.

lbwexler:
Great!  Think this is a great next project for you Yana.  It's been a huge gap for us.  

Let's use the recent 'DockContainer' example as a good example of how to lay it out.  Basically, the summary text with a number of buttons.  Think even two rows of buttons will be fine.  The more examples and variations, the better.

lbwexler:
👍

PR #186, opened 2019-05-14
Popups Example

YanaNirvana:
#174 Added Other > Popups Tab,

Gives many examples of our popups.  
Inspired by the mobile PopupsPage and desktop DockContainerPanel examples.
Includes some examples demonstrating promise from `Alert`s, `Confirm`s, and `Message`s.
Includes examples of anchored and nonanchored `Toast`s


lbwexler:
Think this is fine to merge, but there is an issue with the missing toast icon -- do you have a hoist-react branch with that in it that you create a PR for?

Also, why we are at it, think these should be highlighted a bit more -- they are important!  Let's at least make it first in the 'Other' collection for now....

amcclain:
This looks great - thanks for adding to Toolbox. 👍

PR #192, opened 2019-05-17
Example of short format for relative timestamps

febbraiod:


client-app/src/desktop/tabs/other/RelativeTimestampPanel.js
```diff
@@ -70,7 +74,13 @@ export class RelativeTimestampPanel extends Component {
                             })
                         ]
                     }),
-                    filler()
+                    filler(),
+                    switchInput({
+                        label: 'Short',
+                        labelAlign: 'left',
+                        value: this.useShortFmt,
+                        onChange: v => this.setUseShortFmt(v)
```
febbraiod: We should talk about this. It's working, but there is a weird pattern here, where if you look at the switchInput cmp the onChange doesn't appear to be passed to the cmp (because it's not) but the handler that the cmp calls DOES call the onChange prop. It makes some sense to me as I'm familiar with these inputs, but I could see an app dev being confused by this.

Issue #189, opened 2019-05-15
Grids-> REST Editor: Export to excel is throwing an ERROR


mpeckerman:
1. Navigate to Hoist Reach Toolbox -> Grids tab -> REST Editor panel / screen
2. Click 'download' icon to Export 
3. Error: 'Export not enabled for this grid. See GridModel.enableExport'
https://user-images.githubusercontent.com/40367298/57791598-1ad4a580-770b-11e9-8560-a62bbc78f163.png

PR #194, opened 2019-05-22
`enableExport: true` fix

YanaNirvana:
Bug Fix: Toolbox `REST Grid Panel` now exports appropriately.  Addresses issue #189


TomTirapani:
Hi Yana, Took a quick look at this.

This is great, and certainly fixes the exception reported in the issue. However, to get the most out of the resulting export, I think it's worth letting Excel know that the `earningsDate` column is a date by setting an `ExportFormat` on it (see `hoist-react/cmp/grid/columns` for examples).

This will allow Excel to render the date in a more readable way, as well as enabling things like date sorting, etc.

YanaNirvana:
I'm impressed by your eye for detail, Tom.  That's really impressive.  I didn't even think to check on the format of the export in excel (although I also can't see the excel side of things as I've yet to install the Microsoft office suite on this machine)

PR #198, opened 2019-05-29
Update Toolbox to show off new GridModel.cellBorders

YanaNirvana:
See related **Hoist** Issue [#1041](https://github.com/exhi/hoist-react/issues/1041)


amcclain:
Looks like the addition of the new switch is clipping our row of style controls in SampleGrid - @YanaNirvana can you please take a look?  We can just make the grid container wider to fit them all - they are adding up.

Issue #45, opened 2018-08-22
Example wrapper panel support for "links to source"


amcclain:
Would be useful to pass link(s) to source in Github as props to wrapper panel and have them displayed in a standardized way.  Could be used to link directly to the example source itself (see how it's built) as well as any relevant hoist-react classes.

lbwexler:
+ 1 here on this one.  Would be great to choose which branch of hoist react or toolbox you go to, depending on the the environment...

lbwexler:
Hi Colin -- Can you take a look at this one?  Think its a very important feature, but hopefully not too much work

YanaNirvana:
Discussion with Anselm:

* We want to accept multiple links, the best data structure is perhaps an array of objects, where
* root GitHub URL configurable, use XH.getConf('SourceUrls').toolbox
* support a `note` config in the obj, non-hyperlinked text after the link for additional clarification

amcclain:
Thanks for posting those additional details Yana. 

I think we will find it useful to be able to include links to hoist-react source (the most common, for the components/models we are highlighting), Toolbox source (for an easy link to the source for exactly the module you are viewing), and maybe external links (for things like a link to the ag-grid docs).

Given that, we discussed parsing the URLs passed into the new prop to look for a leading token such as`$TB` or `$HR` - these would be pre-processed to sub in the root from the config, or leave the link alone to support external links.

In terms of display - if any links are configured, the wrapper should write a dedicated links panel into the layout.  I was thinking something that could be docked to e.g. the lower right corner and ideally collapsible.

YanaNirvana:
I don't think we'll get much use out of supporting `$AG/path` since we might have to google for the exact docs (the `path` part of `$AG/path`), and at the same time, I'm happy to build that.

Thanks for giving more details on the display and reminder on the `$TB` and `$HR` :)

YanaNirvana:
collapsible... 🤔How would you feel about using our shiny new docked panel in the bottom right? @amcclain  :D

PR #202, opened 2019-06-04
Wrapper takes in links and link!

YanaNirvana:
Addresses #45:  (edited)

* When wrapper is given `links` prop, it will render a 'Source Code' DockContainer in the bottom right.
* `links` shall be an array of objects or a single object, where:
* the object(s) must be given a `url`, `text` (within hyperlink), and optionally, `notes` (text to the right of hyperlink)
* first draft of docs... requesting suggestions!


client-app/src/desktop/common/Wrapper.js
```diff
@@ -20,10 +54,70 @@ class Wrapper extends Component {
...
+            });
+            
+            return dockContainer({
+                model: dockContainerModel
+            });
+        }
+    }
+    
+    generateLinks() {
+        const arrayLinks = this.props.links || castArray(this.props.link);
```
TomTirapani: See previous comment about supporting either link or links, but not both

client-app/src/desktop/common/Wrapper.js
```diff
@@ -20,10 +54,70 @@ class Wrapper extends Component {
...
+            arrayLinks.map(linkObj => this.renderSingleLink(linkObj))
+        );
+        
+    }
+    
+    renderSingleLink(linkObj) {
+        return p(
+            a({
+                href: this.generateUrl(linkObj.url),
+                item: code(linkObj.text),
```
TomTirapani: While I get what you're going for the monospaced `code` tag here, I think it might look better if we stick with the standard title font :)

client-app/src/desktop/common/Wrapper.scss
```diff
@@ -28,4 +28,8 @@
       }
     }
   }
+
+  .toolbox-wrapper-sourcecode {
+    padding: 10px;
```
TomTirapani: Let's use one of our standard hoist padding vars here, to keep it consistent with any future layout changes (e.g. `padding: var(--xh-pad-px);`

client-app/src/desktop/common/Wrapper.js
```diff
@@ -20,10 +54,70 @@ class Wrapper extends Component {
...
+    renderSourceCodePanel() {
+        const {dockContainerModel} = this;
+        const {links, link} = this.props;
+        
+        if (links || link) {
+            dockContainerModel.addView({
+                id: XH.genId(),
+                icon: Icon.code(),
+                title: code('Source Code'),
+                allowDialog: false,
```
TomTirapani: I think these docked views could be initialised with `collapsed: true`

client-app/src/desktop/common/Wrapper.js
```diff
@@ -20,10 +54,70 @@ class Wrapper extends Component {
...
+    }
+    
+    renderSingleLink(linkObj) {
+        return p(
+            a({
+                href: this.generateUrl(linkObj.url),
+                item: code(linkObj.text),
+                target: '_blank'
+            }),
+            ' | ',
```
TomTirapani: Minor thing, but I think we should only render these separators if the link actually has notes.

client-app/src/desktop/common/Wrapper.js
```diff
@@ -20,10 +54,70 @@ class Wrapper extends Component {
...
             ],
             ...rest
         });
     }
+    
+    renderSourceCodePanel() {
+        const {dockContainerModel} = this;
+        const {links, link} = this.props;
+        
+        if (links || link) {
```
TomTirapani: We could invert the condition here and do an early return of null if there are no links (e.g. `if (!links) return null;`). This allows us to avoid having the bulk of the function in conditional nested params.

It's purely a stylistic choice, but one that I think helps make intentions clearer. There's a nice discussion here on the pros and cons of both styles: https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement

client-app/src/desktop/common/Wrapper.js
```diff
@@ -1,15 +1,49 @@
...
+     *      '$HR' for hoist-react files, such as '$HR/desktop/cmp/button/Button.js'.
+     * @param {string} link.text - text to be displayed in hyperlink.
+     * @param {(string\|React component)} [link.notes] - text to be displayed outside hyperlink.
+     */
+    
+    static propTypes = {
+        
+        link: PT.object,
+        
+        links: PT.arrayOf(PT.object)
```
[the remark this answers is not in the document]
YanaNirvana: Sure thing, it could def save me some logic later

client-app/src/desktop/common/Wrapper.scss
YanaNirvana: good idea!

client-app/src/desktop/common/Wrapper.js
YanaNirvana: 👍

client-app/src/desktop/common/Wrapper.js
YanaNirvana: Oohf.  Thanks for linking the discussion.  I've once had to make something with a lot of error handling, and the `if(somethingInvalid) return null` pattern was a lifesaver.  Much easier to ready than nested if and else's.  

I can see what you mean with inverting the condition -- it allows us to take the bulk of code out of the conditional.  While I don't usually think in this pattern, I agree it's probably good for reading and maintaining later.. 

Thanks again!

client-app/src/desktop/common/Wrapper.js
```diff
@@ -1,15 +1,63 @@
...
+    /**
+     * Wrapper is current standard 'wrapper' for demoing hoist-react components. :)
+     * It accepts:
+     *      (optional) `description` for a quick summary of the component,
+     *      (optional) `links` to source code on GitHub or external websites.
+     *
+     * Note that while `links` is optional and that Developers may choose to provide either
+     * a single Object, or array of Objects, each Object requires a `url` and `text` prop:
+     *
+     * @param {string} link.url - can be a full URL (must include https://) or a string that starts with:
```
lbwexler: We discussed creating a WrapperLink object, as a place to hang data and still think this is a good idea.

client-app/src/desktop/common/Wrapper.js
```diff
@@ -1,15 +1,63 @@
...
+        this.dockContainerModel.addView({
+            id: XH.genId(),
+            icon: Icon.code(),
+            title: code('Source Code'),
+            allowDialog: false,
+            allowClose: false,
+            collapsed: true,
+            content: panel({
+                className: 'toolbox-wrapper-sourcecode',
+                item: this.renderLinks(props)
```
lbwexler: Sorry to make you crazy, but now that you have (correctly!) moved the link generation/view creation *out* of the render() method and into the constructor, these methods should be named createXXX.....

client-app/src/desktop/common/Wrapper.js
```diff
@@ -1,15 +1,63 @@
...
+    static propTypes = {
+        
+        links: PT.oneOfType([PT.arrayOf(PT.object), PT.object])
+    };
+    
+    @managed
+    dockContainerModel = new DockContainerModel();
+    
+    constructor(props) {
+        super();
```
lbwexler: should be `super(props)`

client-app/src/desktop/common/Wrapper.js
```diff
@@ -20,10 +68,42 @@ class Wrapper extends Component {
...
                 }),
-                children
+                children,
+                dockContainer({model: this.dockContainerModel})
             ],
             ...rest
         });
     }
+    
+    renderLinks(props) {
```
lbwexler: once you change to call `super(props)` in the constructor, you won't need to pass the props through as method args like this, they will get installed on the object, and available in all methods....

client-app/src/desktop/common/Wrapper.js
YanaNirvana: I'll start making that

client-app/src/desktop/common/Wrapper.js
YanaNirvana: Regarding line 53, right?

amcclain:
Hi good morning - took a look to see if I can provide any (hopefully useful) input.  Coming right along!

Some quick suggestions:
* The text could be optional.  If not provided, split the url on `/` and take the last entry - this should handle the common case of wanting the file name to be the link text.
* Love the use of DockedViews! I would change the title to be "Source Links" vs. "Source Code" - we might put other kinds of links in there and I want to avoid setting up too much of an immediate expectation that you will expand that panel and be looking at some kind of code editor.
* Need to constraint the width of the links panel so a long note doesn't blow it out.  I suggest always making it a fixed width, around 350-400px maybe?  
* Did you try formatting into an unordered list?  That could let us keep the link and any optional note on the same line which might be a bit more clear / space efficient.  See what it looks like...

Then we can go through and populate these, putting them to use. 👍

client-app/src/common/ToolboxLink.js
```diff
@@ -0,0 +1,54 @@
...
+    createUrl() {
+        const sourceUrls = XH.getConf('sourceUrls');
+        
+        return (this.props.url
+            .replace('$TB', sourceUrls.toolbox)
+            .replace('$HR', sourceUrls.hoistReact)
+        );
+    }
+    
+    createDefaultText() {
```
YanaNirvana: @amcclain 
I took it a step further and wanted to filter out the stuff after `#`, such as when given something like '  I happen to use this a fair bit, so I though it'd be useful.  Open to suggestions :)

YanaNirvana:
Thanks for looking this over & giving suggestions.  

☑️ Text optional 
☑️ "Source Code" is now "Source Links"
☑️ Constrained to 350 px
⏸ Regarding using `<ul></ul>`:  

Screen Shot 2019-06-10 at 3 10 53 PM

I think if we're passing in links of various sizes, it looks messy... what do you think? open to suggestions :)

amcclain:
I see what you mean - how about a table?  Will need to see how it looks in practice, but I think if we had a table with a consistent width - specifically for the first cell with the link - it could look a bit more buttoned up.

We have that style in our readmes and I think it's pretty decent - although I would suggest just making the file name a hyperlink vs. having the dedicated column:

Screen Shot 2019-06-12 at 11 08 39 AM

Issue #208, opened 2019-06-12
Grid switches in a vertical panel on the side


YanaNirvana:
_Originally posted by @lbwexler in https://github.com/exhi/toolbox/pull/205#issuecomment-500959420_

I actually think we need to have those switch options in a little vertical
panel on the side.  Advantages:

+ It will align better.
+ Can add docs,
+ Can vertical scroll (more scalable)
+ Can generally be visually seperated better from the component itself that
is being demonstrated.

PR #209, opened 2019-06-12
Grid options in a side panel

YanaNirvana:
Addresses issue #208: 

Grid now has many options that we want to demo.  These options are now taking up so much space in the bottom toolbar of grids that we had to increase the width on many grids.  Lee suggested that we can move these switches to a side panel.  Here's my interpretation of what that might look like.

Screen Shot 2019-06-12 at 4 30 35 PM
Screen Shot 2019-06-12 at 4 29 16 PM

---

**Alternatively**:

Screen Shot 2019-06-11 at 5 27 09 PM
Screen Shot 2019-06-11 at 5 26 53 PM


TomTirapani:
Overall, this looks great and is a clear improvement! A few thoughts:

- We are currently showing the sample grid in 3 places - `app/grids/standard`, `app/other/mask` and `app/containers/dock` (in the "Complex View"). In the latter two places, the headers are doubled on the sidebar:

Double headers

I think the side panel doesn't need its own header & icon. We can instead just show the resizer bar with the collapse toggle button. This way, we can avoid doubling headers. This will also allow us to revert back to the container being responsible for setting the icon and header here: https://github.com/exhi/toolbox/pull/209/files#diff-4944ad4150bd2a8d9c65437e73c88ee6R27

I think we can revert this, and rely on the outer panel having its own toolbar.

- We can lose the toolbar separators between the switches - now that they are vertically stacked, they are no longer required to visually pair the switches with their labels.

- Not sure what best to do about the "selected row text" in the bottom toolbar - it looks a bit lonely there now! Perhaps we can move it to the bottom of the new side panel? Or just lose it entirely?

amcclain:
Hi - I can take a look at this soon, would love to merge.

@YanaNirvana I agree re. avoiding the doubed-up headers - can/did you tweak to remove those? But also before they go - why is the icon and the title text so squished together on the screenshots above? Is that using the standard props for panel?

The selected item text does look a bit weird, but honestly I wanted to have a very clear example of how to reactively bind to the selection as that's such a key thing to know how to do on grid.  Will think of any other ideas or open of course to suggestions.

YanaNirvana:
@TomTirapani thanks for giving your input.  I think you're right that these look better.  

Screen Shot 2019-06-28 at 10 41 29 AM
Screen Shot 2019-06-28 at 10 41 46 AM
Screen Shot 2019-06-28 at 10 42 00 AM

@amcclain 
Good catch!  The icon and title do get squished together by default, unless we pass in a space in the title string.  I can make a ticket for that if you'd like?

I also like the selected item text & think it's great that we're demoing that.  I think we should keep it, but make its height shorter if possible.  Kind of like how chrome gives a little hint at link URLs when you hover over them (displays in the bottom left, but font size small and unobtrusive).

amcclain:
Playing with this a bit more now and will push some tweaks to merge - thanks!

amcclain:
Hi - merging now, please take a look and of course open to any further suggestions.

Added a little custom header to the collapsible switches panel.  In the past I think we've had some kind of "mini header" component that was useful for when you want a much tighter title for a secondary panel like this. I can imagine us adding something similar to Hoist React here.

Left the other grids with the switches in the toolbar.  Thought about making the GridStyleSwitches component render the collapsible panel and having the UI the same across all grids with these switches available, but I wasn't sure about it.  Can see how the toolbar is still working OK for these - especially the ag-Grid example, and returning a fragment is a clever solution there.

YanaNirvana:
@amcclain Those changes look really nice.  I especially like the improved selbar; it looks posh.  That new little "mini header" looks great too, although I think it might look even better if we increased the contrast between the text and background color.  Maybe white text on the dark blue grey background to match other titles?

PR #223, opened 2019-07-14
Use gridCountLabel in core desktop grid examples

amcclain:
+ Enable multi-select in treeGrid


amcclain:
Supports hoist-react PR w/same branch. Note that new comp could be used in examples, other places - wanted to confirm we were generally happy with it first.

PR #229, opened 2019-07-23
Cube filter fn

febbraiod:
Far from ready for prime time, but want to show this as a POC and open a discussion as to whether or not we think this is worth while. FWIW, my original case has nothing to with any of the value field as with this example, but thought this might be more interesting and informative.

Original need was something like:
const selectedFund = 'Inclusive';
{field: 'fund' filterFn: (val) => { if (selectedFund == 'Inclusive') return true)} }

The value check here is something I think might be more interesting, but also raises the question...Is it too mysterious as it refers to the mktVal on the leaves? (which we're not always showing)


client-app/src/desktop/tabs/forms/InputsPanel.js
```diff
@@ -217,7 +217,8 @@ export class InputsPanel extends Component {
                                 item: dateInput({
                                     showActionsBar: true,
                                     timePrecision: 'minute',
-                                    timePickerProps: {useAmPm: true}
+                                    timePickerProps: {useAmPm: true},
+                                    pickerOnlyMode: true
```
febbraiod: ignore this, somehow snuck in from some other experiment.

lbwexler:
Discussed with Don -- we are going to hold off on this.

PR #230, opened 2019-07-25
Use new IsReady prop on AgGridModel

haynesjm42:
Updated the ag-grid wrapper example to use the new IsReady prop instead of checking the agApi reference


lbwexler:
John -- don't think we want to merge this at this point -- as we discussed, if you are actually using the API, ok to observe the API

PR #231, opened 2019-07-25
TreeMap

TomTirapani:


client-app/src/admin/AppModel.js
```diff
@@ -36,7 +36,12 @@ export class AppModel extends BaseAppModel {
             },
             {
                 name: 'wip',
-                path: '/wip'
+                path: '/wip',
+                children: [
+                    {name: 'simpleTreeMap', path: '/simpleTreeMap'},
+                    {name: 'gridTreeMap', path: '/gridTreeMap'},
+                    {name: 'splitTreeMap', path: '/splitTreeMap'}
```
amcclain: I think we can move these to the main Toolbox app at this point.  Looking for a succinct name for the Charts tab that allows them to live under there... Visualizations?  DataViz?  Leave as charts?

client-app/src/admin/wip/treeMap/GridTreeMapModel.js
```diff
@@ -0,0 +1,96 @@
...
+            },
+            {
+                field: 'mktVal',
+                headerName: 'Mkt Value (m)',
+                headerTooltip: 'Market value (in millions USD)',
+                align: 'right',
+                width: 130,
+                absSort: true,
+                agOptions: {
+                    aggFunc: 'sum'
```
amcclain: What's going on with these? Doesn't portfolioService produce with aggregates?

Also - with this grid and the other splitMap example - I vote to simplify the grid config itself as much as possible since it's not the code we're looking to highlight.  E.g. could drop the colChooser, tooltips, anything else not really essential.  (But still would keep e.g. the number formatting or anything where leaving it out will just look weird or distracting.)

client-app/src/admin/wip/treeMap/GridTreeMapModel.js
```diff
@@ -0,0 +1,96 @@
...
+            {...emptyFlexCol}
+        ]
+    });
+
+    @managed
+    treeMapModel = new TreeMapModel({
+        gridModel: this.gridModel,
+        labelField: 'name',
+        valueField: 'pnl',
+        heatField: 'pnl',
```
amcclain: Can we calc heat as pnl/mktVal or similar to get another dimension in the viz?  (Applies to portfolio example app as well.)

client-app/src/admin/wip/treeMap/GridTreeMapModel.js
```diff
@@ -0,0 +1,96 @@
...
+        valueFieldLabel: 'Pnl'
+    });
+
+    constructor() {
+        // Load data when dimensions change
```
amcclain: We can drop this comment as the reaction speaks for itself here 😄

client-app/src/examples/portfolio/PositionInfoPanel.js
```diff
@@ -0,0 +1,33 @@
+import {Component} from 'react';
+import {elemFactory, HoistComponent} from '@xh/hoist/core';
+import {panel} from '@xh/hoist/desktop/cmp/panel';
+import {hbox} from '@xh/hoist/cmp/layout';
+import {ordersPanel} from './OrdersPanel';
+import {chartsPanel} from './ChartsPanel';
+
+@HoistComponent
+export class PositionInfoPanel extends Component {
```
amcclain: Would suggest `PositionDetailPanel`

client-app/src/examples/portfolio/PortfolioPanelModel.js
```diff
@@ -1,40 +1,54 @@
...
+                    asElement: true
+                })
+            ];
+        },
 
-    @bindable displayedOrderSymbol = '';
+        labelField: 'name',
+        valueField: 'pnl',
+        heatField: 'pnl',
+        valueFieldLabel: 'Pnl',
```
amcclain: "P&L" please (my OCD always gets tripped up by this one).

We might want to change the default tooltip value format (within the map code) to use formatLedger as well - I think that will be what we want most of the time.

PR #236, opened 2019-07-29
Added Select with selectOnFocus = true

cnrudd:


amcclain:
Thanks for setting this up - I ended up just adding that config + enableCreate to two existing selects so we could ensure we had it in there w/o adding another to this already cramped layout.

We need to get that panel under control one day soon - it's completely crazy.

PR #249, opened 2019-08-08
Fix value binding to mobile toolbar select example.

### LINKED ISSUE #246: Select won't take value in mobile toolbars example

TomTirapani:
Fixes #246


amcclain:
Great thank you - figured this was exactly it.  Even if it seems like a bit of overkill, avoids confusion over the issue (is the component broken? does it somehow not really work in toolbars?) should someone stumble across it.

PR #256, opened 2019-08-29
Chart instance access

amcclain:
WIP pending the larger refactors detailed in https://github.com/exhi/hoist-react/pull/1329 - but would like to keep this branch open and setting up this PR to track so we can use the (minimal) UI updates here to test the revised approach.


lbwexler:
Merging  this minor very old PR to use the chartInstance now available from Hoist

PR #261, opened 2019-09-06
fix check/uncheck on tree grid checkboxes

cnrudd:


lbwexler:
We should go over this.  Would be nice to improve support for this in hoist if needed

lbwexler:
Fixed this in a refactor here: https://github.com/xh/toolbox/commit/3257e77b2eb7daff5b32a0dfa237f6f3f7edd087

PR #277, opened 2019-10-02
Store/Record API updates, Store editing test panel

haynesjm42:
Updates for Store/Record API changes, Store editing test panel


client-app/src/desktop/common/SampleTreeGridModel.js
```diff
@@ -55,15 +59,26 @@ export class SampleTreeGridModel {
...
+    async doLoadAsync({isRefresh, isAutoRefresh}) {
         const {gridModel, dimChooserModel} = this,
             dims = dimChooserModel.value;
 
         return XH.portfolioService
             .getPositionsAsync(dims, true)
             .then(data => {
-                gridModel.loadData(data);
-                gridModel.selectFirst();
+                if (isRefresh || isAutoRefresh) {
```
lbwexler: just FYI -- should be enough to just check if isRefresh is true.  `isAutoRefresh` implies `isRefresh`

PR #278, opened 2019-10-02
Cleanup tree w/checkbox tree column renderer

haynesjm42:
Changes for https://github.com/xh/hoist-react/pull/1392


lbwexler:
This looks like a great cleanup.

What about the line here:  Is it needed?
        props.reactContainer.style = 'display: inline-block';

PR #281, opened 2019-10-10
Fetch307

cnrudd:
This adds a test page for Hoist-React fetchservice to test how it handles all status codes.  We know it handles the common ones 200, 204, 401, 404, 500, but there are many more.

In particular, this test page lets us see how fetchservice handles 307 redirect status codes.
Issue #1347 in hoist-react https://github.com/xh/hoist-react/issues/1347 suggested that fetchservice did not handle redirect responses correctly.

But I think that it does, if the endpoint redirected to responds with the same content-type that the original request was expecting:  
 json -> redirect -> json   : all good
 json -> redirect -> html :  parse json throws error

### DIFF
--- a/client-app/src/admin/tests/TestsTab.js
+++ b/client-app/src/admin/tests/TestsTab.js
@@ -13,6 +13,7 @@ import {CubeDataPanel} from './cube/CubeDataPanel';
 import {WebSocketTestPanel} from './websocket/WebSocketTestPanel';
 import {LocalDateTestPanel} from './localDate/LocalDateTestPanel';
 import {PanelResizingTestPanel} from './panels/PanelResizingTestPanel';
+import { FetchApiTestPanel } from './fetch/FetchApiTestPanel';
 
 
 export const TestsTab = hoistCmp(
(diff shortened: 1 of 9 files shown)


client-app/src/admin/tests/TestsTab.js
```diff
@@ -13,6 +13,7 @@ import {CubeDataPanel} from './cube/CubeDataPanel';
 import {WebSocketTestPanel} from './websocket/WebSocketTestPanel';
 import {LocalDateTestPanel} from './localDate/LocalDateTestPanel';
 import {PanelResizingTestPanel} from './panels/PanelResizingTestPanel';
+import { FetchApiTestPanel } from './fetch/FetchApiTestPanel';
```
lbwexler: inconsistent style

client-app/src/admin/tests/fetch/FetchApiTestPanel.js
```diff
@@ -0,0 +1,90 @@
...
+import { panel } from '@xh/hoist/desktop/cmp/panel';
+import { select, jsonInput} from '@xh/hoist/desktop/cmp/input';
+
+import { FetchApiTestModel } from './FetchApiTestModel';
+
+import './FetchApiTestStyles.scss';
+
+
+export const FetchApiTestPanel = hoistCmp({
+    model: creates(() => new FetchApiTestModel()),
```
lbwexler: can be just creates(FetchApiTestModel)

client-app/src/admin/tests/fetch/FetchApiTestModel.js
```diff
@@ -0,0 +1,413 @@
...
+                    console.log(resp);
+                    
+                    this.setResponse(JSON.stringify(resp, undefined, 2));
+                }, 
+                (err) => this.handleError(err)
+            );
+        }
+    }
+
+    async doFetchASync(code) {
```
lbwexler: typo here in method name (ASync).  Curious about use of switch here, and the use of 2 arguments to then(), rather than catch

lbwexler:
Colin and I discussed some changes to run this in bulk that he is going to look at.

Issue #263, opened 2019-09-13
FileManager Example not updating after deleting file on server


lbwexler:
- Add a file
- Sync it to the server
- delete it 
- sync the delete to the server

...And then you need to refresh the page to actually see the file gone.....

peterjcoffman:
Fixed by https://github.com/xh/toolbox/pull/287

PR #287, opened 2019-11-21
Fix FileManager deletion (fixes: #263)

peterjcoffman:


lbwexler:
Congrats on your first PR with XH!

👍

amcclain:
Excellent, thank you for fixing what was surely my mistake!

Not returning those promises is a common error. Note that in Groovy you get an implicit return of the last statement's output, but certainly not the case in JS.  (And even in Groovy we don't typically rely on it unless we're writing very terse one-liners.)

Issue #247, opened 2019-08-07
Portfolio example - configure multi-level tree as default


amcclain:
Currently I _believe_ the dimension chooser is unconfigured, causing it to default to a one-level "tree" with the first dimension in the list (Fund) - which happens to be the most boring grouping you could show out of the box in terms of row count and failing to show off the entire tree idea.

Suggest Sector > Symbol as a better default grouping. We could also look at configuring the grid to expand all those nodes by default.

Note that I do *not* think we should wire up a preference here to remember the last grouping, as multiple random users might change and leave it on a setting we don't care to show off - we want to control that default / first look at the example for new viewers.

As part of this, would like to pre-config a few other useful groupings in the dimension chooser's history list.  I think that is possible.

peterjcoffman:
There is currently a preference to remember previous groupings. However, it is a local preference so it will not affect other users.

peterjcoffman:
Pull request is here:
https://github.com/xh/toolbox/pull/288

PR #288, opened 2019-11-21
Portfolio defaults

peterjcoffman:
The tree in the Portfolio example app will now be expanded by default and will display a more interesting grouping.

Will fix issue https://github.com/xh/toolbox/issues/247

### DIFF
--- a/client-app/src/examples/portfolio/GridPanelModel.js
+++ b/client-app/src/examples/portfolio/GridPanelModel.js
@@ -11,7 +11,7 @@ export class GridPanelModel {
     @managed panelSizingModel = new PanelModel({
         defaultSize: 500,
         side: 'left',
-        prefName: 'portfolioMapPanelConfig'
+        prefName: 'portfolioGridPanelConfig'
     });
 
     @bindable loadTimestamp;
--- a/client-app/src/examples/portfolio/GridPanelModel.js
+++ b/client-app/src/examples/portfolio/GridPanelModel.js
@@ -2,11 +2,17 @@ import {HoistModel, XH, managed, LoadSupport} from '@xh/hoist/core';
 import {bindable} from '@xh/hoist/mobx';
 import {GridModel} from '@xh/hoist/cmp/grid';
 import {fmtNumberTooltip, millionsRenderer, numberRenderer} from '@xh/hoist/format';
-
+import {PanelModel} from '@xh/hoist/desktop/cmp/panel';
 
 @HoistModel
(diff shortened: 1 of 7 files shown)


amcclain:
Hi - thanks for opening this PR Peter - looks good.

When checking it out, I realized there is another unrelated-but-right-nearby issue that I’ve caught a few times and is frustrating.  It relates to some limitations on our panel sizing model.  If you open up the app with a relatively narrow window, the grid can end up being very constrained width-wise and needs to scroll horizontally.  Sometimes you need to accept horiz scrolling when you have very wide grids, but it’s silly for this grid which has only three columns and is the most important dataset in the app.

The issue is that PanelModel needs to take a default size, in pixels, for the side that you wish to be collapsible. We have the map as collapsible and are giving it a default size of 1000px, meaning that any window size close to that (which is still pretty reasonable) doesn’t give the grid much or any room. Conversely, if you open the app on a very wide monitor, the grid has much *more* room than it needs.

I think a better default for this particular app / example - at this point in the evolution of things - would be to remove the resizability between those two top panels.  Give the grid a fixed width that is likely to display the data at reasonable levels of grouping and let the map flex to soak up the rest.

Thoughts? Play with it and see what I mean.  I'd like to include any updates in this PR - they aren't directly related so they don't *need* to go together, but they share a common goal of ensuring this example makes the best first impression possible.

PS - improving Panel's sizing model and allowing for more / better layout options is on the roadmap, potentially along with the use of a layout management library. So I don't think we want to do much of a deep-dive there right this moment.

amcclain:
Edit - a better idea I think:  we could flip it and make the grid the thing that collapses away - so you could collapse it to see only the map.  That lets us still provide the resizability but we can provide the grid side with a default pixel width that works for us.  

In the end I think it makes much more sense to collapse the grid and see only the map - you can always fit the grid into view so making it full-width doesn't add much. But getting it out of the way so you have more/only map and can play around with that visualization more makes sense.

peterjcoffman:
I believe these are all the code changes needed. The defaults I chose seem sensible to me, but feel free to change them.

As I added new default preferences to bootstrap.groovy, applying these changes will require redeploying the Grails app.

lbwexler:
reviewed this with peter.  He will clean up a few minor nits.

amcclain:
Great thanks! On my phone now but will take a look and merge this in once properly back at my desk. 

Appreciate your going the extra step here to tune this screen up.

Issue #244, opened 2019-08-07
Status monitors for new Grails services


amcclain:
Now that we have some nice server-side services, we can write some meaningful status monitors within `MonitorDefinitionService` and provide a much better demo of that feature.

peterjcoffman:
There are a few things in the existing monitors that need to be changed. Most notably, we get NullPointerExceptions if the news has never been fetched. Also, the threshold on Loaded Stories is incorrect. It's set to zero, which means it will only fail if there are fewer than zero stories (clearly impossible). I'll fix those things while I'm adding more monitors.

I also think there's a shortcoming in the way the MonitorService currently works. Let's say you want the monitor to fail if the most recent news story was over 360 minutes ago, like it works now. If the last news story does not exist, i.e. the news has NEVER been fetched, there is no sensible way to handle that. I think the way to do it would be to "abort" using the numeric metric and instead just print a message and manually set it to fail.

We could do that by, say, having MonitorResultService.evaluateThresholds do nothing at all if the resultStatus is already set.

peterjcoffman:
I don't think that change is important. We can come pretty close to the behavior I want within the current API.

PR #289, opened 2019-11-22
New monitors

peterjcoffman:
Adds several new monitors to toolbox, including storage space used, presence of any "malicious files" (i.e. files with names containing .sh or .exe), status of connection to FDA website, and server memory usage.

Also corrects a few edge cases in the existing monitors where they could behave unexpectedly if the server has not yet fetched any news stories.

Implements new monitors per https://github.com/xh/toolbox/issues/244

### DIFF
--- a/grails-app/init/io/xh/toolbox/BootStrap.groovy
+++ b/grails-app/init/io/xh/toolbox/BootStrap.groovy
@@ -238,6 +238,15 @@ class BootStrap {
                 metricUnit: '',
                 active: true
         )
+        createMonitorIfNeeded(
+                code: 'memoryUsage',
+                name: "Memory Usage of Server",
+                metricType: 'Ceil',
+                metricUnit: '%',
+                warnThreshold: 40,
+                failThreshold: 75,
+                active: true
+        )
     }
 
     private void createMonitorIfNeeded(Map data) {
--- a/grails-app/init/io/xh/toolbox/BootStrap.groovy
+++ b/grails-app/init/io/xh/toolbox/BootStrap.groovy
@@ -189,22 +189,22 @@ class BootStrap {
 
     private void ensureMonitorsCreated() {
         createMonitorIfNeeded(
-                code: 'instrumentCount',
-                name: 'Number of Instruments',
-                metricType: 'Floor',
-                metricUnit: 'instruments',
-                warnThreshold: 50,
-                failThreshold: 10,
-                active: true
+            code: 'instrumentCount',
+            name: 'Number of Instruments',
+            metricType: 'Floor',
+            metricUnit: 'instruments',
+            warnThreshold: 50,
+            failThreshold: 10,
+            active: true
         )
         createMonitorIfNeeded(
-                code: 'positionCount',
-                name: 'Number of Positions',
-                metricType: 'Floor',
-                metricUnit: 'positions',
-                warnThreshold: 50,
-                failThreshold: 10,
-                active: true
+            code: 'positionCount',
+            name: 'Number of Positions',
+            metricType: 'Floor',
+            metricUnit: 'positions',
+            warnThreshold: 50,
+            failThreshold: 10,
+            active: true
         )
         createMonitorIfNeeded(
             code: 'newsStoryCount',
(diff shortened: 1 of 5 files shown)


grails-app/init/io/xh/toolbox/BootStrap.groovy
```diff
@@ -212,6 +213,40 @@ class BootStrap {
...
+                metricUnit: 'MiB',
+                warnThreshold: 16,
+                failThreshold: 32,
+                active: true
+        )
+        createMonitorIfNeeded(
+                code: 'maliciousFilesFound',
+                name: 'Malicious Files Uploaded',
+                metricType: 'Ceil',
+                metricUnit: 'potentially malicious files',
```
lbwexler: 'files' suffices for unit

grails-app/init/io/xh/toolbox/BootStrap.groovy
```diff
@@ -212,6 +213,40 @@ class BootStrap {
...
+                warnThreshold: 0,
+                failThreshold: 1,
+                active: true
+        )
+        createMonitorIfNeeded(
+                code: 'recallsFetchStatus',
+                name: "Connection status to ${configService.getString('recallsHost', 'api.fda.gov')}",
```
lbwexler: Would just say FDA here for the monitor name -- this name is getting fixed in the DB, and this code mistakenly implies that it is dynamic,

grails-app/init/io/xh/toolbox/BootStrap.groovy
```diff
@@ -212,6 +213,40 @@ class BootStrap {
             metricUnit: 'sources',
             active: true
         )
+        createMonitorIfNeeded(
+                code: 'storageSpaceUsed',
+                name: 'Storage Space Used by Uploaded Files',
```
lbwexler: '....Used by File Manager Example App'

grails-app/services/io/xh/toolbox/MonitorDefinitionService.groovy
```diff
@@ -5,25 +5,103 @@ import io.xh.hoist.monitor.MonitorResult
...
 
 class MonitorDefinitionService extends BaseService {
 
     def newsService
+    def fileManagerService
+    def recallsService
 
+    /**
+     * Check the count of news stories loaded by NewsService
+     * @param result
```
lbwexler: Leave the boiler plate (and unfilled in) @param and @return out here.  These will be the same for every monitor.

Looking for minimal boiler plate in our examples.

grails-app/services/io/xh/toolbox/MonitorDefinitionService.groovy
```diff
@@ -5,25 +5,103 @@ import io.xh.hoist.monitor.MonitorResult
...
+        //sum up the sizes of all uploaded files..
+        def bytes = fileManagerService.list()
+                .collect {it.length()}
+                .sum() ?: 0
+        //and convert to megabytes rounded to two decimal places
+        def MiB = (bytes / (1024 * 1024)).setScale(2, BigDecimal.ROUND_HALF_UP)
+        result.metric = MiB
+    }
+
+    /**
```
lbwexler: I am not a fan of this monitor.  Think it raises more questions then we want to.

grails-app/services/io/xh/toolbox/app/RecallsService.groovy
```diff
@@ -7,7 +7,17 @@ class RecallsService extends BaseService {
 
     def configService
 
+    //last time we tried to fetch data, did we successfully connect to the server?
```
lbwexler: commenting style  is to have a space after the //.  

also, these comments are probably superfluous -- will discuss our philosophy here in person

amcclain:
Thanks - these are coming along nicely and great to see some additional monitor patterns / examples that we can demo with different kinds of metrics.

Would like to write 2-3 that called/exercised the PortfolioService and related - these we can sort to the top of the list and will look the most like monitors we might write for a client's application. I am thinking of things like total position count, security (instrument) count, last pricing update - stuff that's pretty easy to code up but which you can imagine actually monitoring in a portfolio dashboard style app.

peterjcoffman:
@amcclain I can certainly make those monitors, but what would be a sensible metric? When would I warn about the number of positions or the number of instruments?

lbwexler:
Its an arbirtary low number.  Could be 1 or 10 or 100 for the bootstrap. 

We don't expect it to ever actually fire with our synthetic data, it is for show. 

In a real production system, it would be clear that a drastically reduced number like that is indicating a problem.

peterjcoffman:
@lbwexler So it should fail if it's too low, not if it's too high?

lbwexler:
yes!

peterjcoffman:
We update pricing every day. I don't believe there's a more granular time than that.

lbwexler:
Peter and I reviewed in depth.

Issue #293, opened 2019-11-29
Correct OHLC spelling


peterjcoffman:
We sometimes use the term "OLHC". The chart is actually "OHLC", so this should be fixed.

peterjcoffman:
PR is here:
https://github.com/xh/toolbox/pull/294

PR #294, opened 2019-11-29
Rename all references to 'olhc'

peterjcoffman:
Should fix this issue:
https://github.com/xh/toolbox/issues/293


amcclain:
Thank you for this! I had noticed this before and meant to file/fix - it's the kind of detail we want to get right - and is somewhat embarrassing to get wrong. 👍

PR #295, opened 2019-12-02
Correct olhc

peterjcoffman:
I missed a few OLHC's last time around. This should get them all.


peterjcoffman:
Should be good to go, I think.

PR #305, opened 2019-12-17
Add check that grid is ready before selecting row

peterjcoffman:
Relates to this issue:
https://github.com/xh/hoist-react/issues/1483

### DIFF
--- a/client-app/src/desktop/common/SampleGridModel.js
+++ b/client-app/src/desktop/common/SampleGridModel.js
@@ -192,8 +192,7 @@ export class SampleGridModel {
             gridModel = this.gridModel;
 
         gridModel.loadData(trades, summary);
-        // if agApi is null, grid is not ready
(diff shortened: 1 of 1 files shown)


client-app/src/desktop/common/SampleGridModel.js
```diff
@@ -192,7 +192,8 @@ export class SampleGridModel {
             gridModel = this.gridModel;
 
         gridModel.loadData(trades, summary);
-        if (!gridModel.hasSelection) gridModel.selectFirst();
+        // if agApi is null, grid is not ready
+        if (gridModel.agApi && !gridModel.hasSelection) gridModel.selectFirst();
     }
```
lbwexler: This is admittedly a very weird bit of code to have to write to avoid the warning.  We have discussed adding a public getter of the form grid.isReady(). In any case, comment is more verbose than needed for this code...

PR #308, opened 2019-12-18
Add tooltip element renderer example to sample grid

haynesjm42:
Example for https://github.com/xh/hoist-react/pull/1485

### DIFF
--- a/client-app/src/desktop/common/SampleGrid.scss
+++ b/client-app/src/desktop/common/SampleGrid.scss
@@ -6,6 +6,36 @@
     background-color: var(--xh-bg-highlight);
     align-items: center;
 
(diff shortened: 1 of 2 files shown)


amcclain:
Hi - sampleGrid still has a number of specific `tooltip` configs on the individual columns, which are now confusing as they are replaced by the (very pretty!) new row-level tooltip element.

@haynesjm42 and @lbwexler let's touch base and discuss briefly - I think it might be best to allow sampleGrid to stick with the more vanilla field-wise tooltips and demo this more advanced pattern somewhere else within Toolbox, in the interest of making sampleGrid as broadly applicable as possible. 

But if we want to keep this example here, let's clean up the column-level tooltips for clarity.

PR #313, opened 2019-12-30
Examples usages of GridModel.ensureSelectionVisible()

peterjcoffman:
Adds a scroll button to the existing Grid Performance test page.

Also renames the Grid Performance tab to just Grid, since it relates to more than just performance.

### DIFF
--- a/client-app/src/desktop/tabs/inputs/tabs/TimeInputPanel.js
+++ b/client-app/src/desktop/tabs/inputs/tabs/TimeInputPanel.js
@@ -1,0 +1,70 @@
+import {useLocalModel} from '@xh/hoist/core/hooks';
+import {InputTestModel} from '../InputTestModel';
+import {PropTypes as T} from 'react-view';
+import {TimeInput} from '@xh/hoist/desktop/cmp/input';
+import {Icon} from '@xh/hoist/icon';
+import {hoistCmp} from '@xh/hoist/core';
+import {inputTestPanel} from '../InputTestPanel';
+
+export const TimeInputPanel = hoistCmp({
+
+    render() {
+        const model = useLocalModel(createModel);
+        return inputTestPanel({model});
+    }
+});
+
+function createModel() {
+    return new InputTestModel({
+        description:
+            [
+            ],
+        componentName: 'TimeInput',
+        props: {
+            /** Initial time the TimePicker will display. This should not be set if value is set. */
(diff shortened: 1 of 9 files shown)


client-app/src/desktop/tabs/inputs/tabs/TimeInputPanel.js
```diff
@@ -0,0 +1,70 @@
+import {useLocalModel} from '@xh/hoist/core/hooks';
+import {InputTestModel} from '../InputTestModel';
+import {PropTypes as T} from 'react-view';
+import {TimeInput} from '@xh/hoist/desktop/cmp/input';
+import {Icon} from '@xh/hoist/icon';
+import {hoistCmp} from '@xh/hoist/core';
+import {inputTestPanel} from '../InputTestPanel';
+
+export const TimeInputPanel = hoistCmp({
```
amcclain: This file does not look like it should be included in this pull request. Was this branch created directly off of develop?

lbwexler:
On the first point -- I discussed this with Peter.  We planned to abandon
that PR, and he was preparing another branch that places a test of the new
functionality in the admin grid testing area.

-Lee

peterjcoffman:
> How about placing the button in the "Display Options" fly-out for the sample grids? That is already clearly set aside as a Toolbox-specific set of example controls, different from anything you would have in an app. I think we can even keep the title of that panel the same - scroll-to-selected is closely enough related to "display" IMO to leave that title in place.

That's not entirely straightforward. The 'Display Options' fly-out expects an AgGridModel, not a GridModel. We use that fly-out in places where no GridModel even exists, such as the ag-Grid Wrapper.

peterjcoffman:
>let's have a quick discussion first to see if we want to do anything with a deeper integration here, or just call the method in a one-off reaction within the Portfolio example.

It would be really trivial to add this behavior to the TreeMapModel. Just add 
```        this.gridModel.ensureSelectionVisible();```
to the bottom of the defaultOnClick method.

lbwexler:
Let's just remove it from the public area, so we can merge the button in the test area and move on.  Would much rather we work on adding this behavior to TreeMapModel now.

lbwexler:
please create a new ticket in hoist-react for the treeMap work and get started on it.
(Think we will want to do it on syncSelection() rather than the click handler, but will let you take a look)

PR #317, opened 2020-01-10
Add a new tab to Admin tests panel for Selects

peterjcoffman:
A demonstration of every currently-supported form of Select.


client-app/src/admin/tests/Select/SelectTestPanel.js
```diff
@@ -0,0 +1,200 @@
...
+            )
+        ],
+        alignItems: 'center'
+    })
+);
+
+@HoistModel
+class LocalModel {
+    @bindable
+    Select;
```
lbwexler: If this represents a value it should have a name like selectValue.  These look like class names.

client-app/src/admin/tests/Select/SelectTestPanel.js
```diff
@@ -0,0 +1,200 @@
...
+    @bindable
+    CreatableSelect;
+    @bindable
+    AsyncSelect;
+    @bindable
+    AsyncCreatableSelect;
+    @bindable
+    BigSelect;
+    @bindable
+    NumOptions = 1000;
```
lbwexler: We don't use this kind of casing for properties in javascript.

client-app/src/admin/tests/Select/SelectTestPanel.js
```diff
@@ -0,0 +1,200 @@
...
+    const numOptionsName = (enableWindowed ? 'Windowed' : '') + 'NumOptions';
+    return fragment(
+        p((enableWindowed ? 'Windowed' : '') + 'Select with many options'),
+        label('value: ' + model[name]),
+        select({
+            model,
+            enableWindowed,
+            bind: name,
+            options: function() {
+                let options = [];
```
lbwexler: generate these in the model itself, with a reaction.

client-app/src/admin/tests/Select/SelectTestPanel.js
```diff
@@ -0,0 +1,200 @@
...
+                    exampleAsync(model, true),
+                    exampleAsyncCreatable(model, true),
+                    exampleBig(model, true)
+                )
+            )
+        });
+    }
+});
+
+function exampleSelect(model, enableWindowed) {
```
lbwexler: these should be hoist components -- let's discuss why.

lbwexler:
Peter -- marked this as WIP.  Let's take a look at this together on Monday -- think we can whip the code into shape pretty quickly.  Also, would like to have our tests refer to our options, (e.g. enableCreate, asyncFn) rather than the names of the underlying react-select classes.

peterjcoffman:
@lbwexler 

I think I agree with you on all points. I was trying to be too lazy.

I've pushed a new version.

PR #324, opened 2020-01-16
Add rangeSelector and default zoom to charts

peterjcoffman:
I added a range selector to both charts with a default 3-month zoom.

I also added:
-  a navigator to the OHLC chart for consistency
-  left margins so that the charts will keep consistent size with changing labels

And I removed some options from the OHLC chart that had no effect.


client-app/src/examples/portfolio/detail/OHLCChartModel.js
```diff
@@ -13,7 +13,8 @@ export class OHLCChartModel {
     constructor() {
         this.addReaction({
             track: () => this.symbol,
-            run: () => this.loadAsync()
+            run: () => this.loadAsync(),
+            fireImmediately: true
```
lbwexler: why is this fireImmediately needed?

peterjcoffman: It isn't needed. I was trying to fix the somewhat flaky loading.

As you keep saying, I should look over the diff before opening a PR.

PR #331, opened 2020-01-24
Move itemHeight to model

peterjcoffman:
We will actually want to merge this one.


TomTirapani:
There are more instances of DataViews in toolbox that will need to be updated in this PR. There's the news example, and one in the mobile app

PR #334, opened 2020-01-27
Modify to reflect changes in Record API

peterjcoffman:


client-app/src/examples/news/NewsPanel.js
```diff
@@ -48,5 +48,7 @@ export const newsPanel = hoistCmp.factory({
 });
 
 function onRowDoubleClicked(e) {
-    if (e.data.url) window.open(e.data.url, '_blank');
+    if (e.data.data.url) {
+        window.open(e.data.data.url, '_blank');
```
lbwexler: Your code looks right -- bit ugly with the doubling up of 'data'  here -- too bad

PR #335, opened 2020-01-28
Minor Top-level tab re-org

amcclain:
+ Re-ordered top-level tabs to put grid / panel before less interesting containers example
+ Moved Formats under Other
+ Tweaks to a few layouts for improved / more consistent sizing


lbwexler:
Nice cleanup!  Agree with highlighting more higher-level and rich components.

PR #339, opened 2020-01-28
Switch idSpec

peterjcoffman:
Use XH.genId to ensure all news items will have a unique ID


peterjcoffman:
I modified the Grails server to give 20 copies of the same news story and this works fine:

image

PR #342, opened 2020-01-30
Remove duplicate updates

peterjcoffman:
`allDescendants` already includes the parent. Therefore, we do not need to explicitly update `rec`; updating `allDescendants` will catch `rec` at the same time.

We also made an incorrect call to modifyRecords. It expects to receive an object with the ID to upgrade.

### DIFF
--- a/client-app/src/desktop/common/grid/SampleTreeGridModel.js
+++ b/client-app/src/desktop/common/grid/SampleTreeGridModel.js
@@ -189,7 +189,7 @@ export class SampleTreeGridModel {
             ];
 
         store.modifyRecords(updates);
-        rec.forEachAncestor(it => store.modifyRecords(it, {isChecked: calcAggState(it)}));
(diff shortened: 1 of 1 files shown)


client-app/src/desktop/common/grid/SampleTreeGridModel.js
```diff
@@ -185,7 +185,6 @@ export class SampleTreeGridModel {
         const {store} = this,
             isChecked = !rec.data.isChecked,
             updates = [
```
lbwexler: seems to me, we might need the first bit for a leaf node, no?  Did you try checking a leaf node ???  :-)

haynesjm42: I think this is ok because our `allDescendants` will always return the record, even if it has no children (if I am reading the code correctly).

lbwexler: great -- need to start reading the entire file -- sorry for muddying the waters here.

PR #343, opened 2020-01-30
React view demo

peterjcoffman:
I think this thing is, at long last, just about done.


client-app/webpack.config.js
```diff
@@ -1,13 +1,16 @@
...
+        ...configureWebpack({
+            appCode: 'toolbox',
+            appName: 'Toolbox',
+            appVersion: env.appVersion || '2.0-SNAPSHOT',
+            agGridLicenseKey: 'Extreme_Heavy_Industries__Scout_3Devs3_April_2020__MTU4NTg2ODQwMDAwMA==8f836aa25989abb0db36231c347dd436',
+            favicon: './public/favicon.png',
+            devServerOpenPage: 'app/',
+            ...env
+        }),
+        node: {fs: 'empty'}
```
peterjcoffman: This is, admittedly, a bit of a hack. It exists because babel tries to import 'fs' (but not actually use it), and this turns that into a no-op.

amcclain: This is a pretty standard approach we have used on client projects for the same reason - we could consider baking it into hosit-dev-utils.

lbwexler:
Closing.  At this point, this is quite obsolete.

Issue #348, opened 2020-02-06
Use better default password for toolbox


lbwexler:
We want to remove the default auth for the non-admin user on toolbox, but in the meantime would like to "complexify" our default demo password.  Will be more professional, and hopefully avoid warnings from chrome.

PR #349, opened 2020-02-06
Changed demo password for toolbox

petradish:
This addresses toolbox ticket: #348 
- Changes demo password on Bootstrap.
- Displays new password to client on toolbox app's login panel message.

### DIFF
--- a/client-app/src/apps/app.js
+++ b/client-app/src/apps/app.js
@@ -14,5 +14,5 @@ XH.renderApp({
     webSocketsEnabled: true,
     idleDetectionEnabled: true,
     checkAccess: 'APP_READER',
-    loginMessage: '👤 toolbox@xh.io + 🔑 toolbox'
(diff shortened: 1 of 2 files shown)


amcclain:
Hi - thanks for jumping on this. 

The one issue we have here is that the Bootstrap user is created within a `createUserIfNeeded()` method - which means that this code will not run if the user is already there, which it is in dev/prod.

I think we should add an additional line to always set the password on the user - after we've created it (if needed) we can then get it again, set that password, and save the user. Does that make sense? 

After this update, it will be very interesting to see how long it takes Google to potentially flag it again. 

We should also prioritize updating TB in prod ASAP. I am guessing that we'll need to release Hoist first - TB _most likely_ has changes that expect the latest Hoist code.  Would be interesting to check.  Assuming so, could you please prep another PR with the same exact change against master so we can do what we would call a "hotfix release"?

amcclain:
Closing this in favor of #350

PR #350, opened 2020-02-06
Changed demo password for toolbox

petradish:
This addresses toolbox ticket: #348

- Changes demo password on Bootstrap.
- Sets passwords on all users on server start
- Displays new password to client on toolbox app's login panel message.


amcclain:
I'm going to close this one as well as (happily) we got to the point with hoist-react where we were ready to do a Toolbox release anyway.  Will re-open against develop with this same branch and then merge. 

👍

PR #354, opened 2020-02-10
Fix data views in Toolbox

peterjcoffman:


amcclain:
@lbwexler this can be used for testing https://github.com/xh/hoist-react/pull/1618 and should be merged along with it, although the News example needs to be tuned up before we release.

(The style names there are not to standard, layout not quite right, and the entire thing seems more complicated than it needs to be and is not currently a good example.)

Issue #23, opened 2018-07-12
Additional "Other" tab ideas


lbwexler:
* **Other > Preferences** tab to demonstrate use of prefService on client. Could include appropriate controls to set / toggle various fun example preferences of the different available data types, storing the updated preference on change. The overall description of the tab could provide a very high-level overview of preferences and prompt the user to reload the page or inspect local storage to see their pref. values.
* **Other > Exceptions** tab with buttons that will generate and catchDefault an exception with various combinations of flags (e.g. showAsError, requireReload, etc.).  Smoke tester can run through buttons quickly.
* **Other > Other Actions** tab (??) that could have buttons to do other programmatic things in framework such as:
  * Call `idleService.suspendApp()` to show "app is sleeping" feature
  * Call `XH.getUser()` and display details

lbwexler:
Happy to take any low hanging fruit here and close this bug.  Doesn't have to be perfect. 

Would be cool to show off bringing up the AppUpdate banner (in a simulated way) as well.

lbwexler:
Think we might put much of this stuff in a top-level tab called "App Lifecycle" or App Support":
Sub-tabs could be:

+ Identity Management (maybe discuss impersonation, which we can't show?)
+ Preferences
+ Exception Handling
+ Update Banner
+ Application Idle

amcclain:
Still agree some good examples to be added re. preferences and (especially) exceptions, which have a clear UI component and are a great example of some Hoist-level polish that apps should leverage.

The preference example could allow the user to set a few preferences via bound form controls - e.g. a boolean (switchInput), string (I'd suggest a select), and JSON (jsonInput). All three should be set to local so they are persisted only to the user's own browser.

Exception example could, obviously, trigger an exception, to show the handler dialog. Only trick there is that we would not want to log on the server or trigger an email to our support list.

lbwexler:
Would be great to get our websocket support in there.  Maybe the ability to send a message via a websocket, and get the server to answer?  Could display the json in a little text box...

"Hoist, tell me a Dad Joke?"

lbwexler:
Nothing particular interesting here, that we wouldn't want a fresh take on if we revisit.  Closing this ancient ticket

PR #357, opened 2020-02-14
Added App Notification examples to Other tab

petradish:
This PR addresses part of issue https://github.com/xh/toolbox/issues/23
Created "App Notifications" tab under Other tab to showcase:
- App Update Banner
- Idle Service
I thought these two tiles made sense as a grouping of notifications from the server.

Will submit separate PRs for Exception Handler, Preferences, and getUser


amcclain:
I think this looks great and will merge - made some small tweaks to the language.

We can continue to re-org this Other tab as needed while adding more examples like this.

PR #360, opened 2020-02-20
Use New Debounce

lbwexler:


client-app/src/examples/recalls/RecallsPanelModel.js
```diff
@@ -104,7 +104,7 @@ export class RecallsPanelModel {
         this.addReaction({
             track: () => this.searchQuery,
             run: () => this.loadAsync(),
-            delay: ONE_SECOND
+            debounce: ONE_SECOND
```
amcclain: Agree this would be a case where we wouldn't want to use the shorter/leading default I proposed in https://github.com/xh/hoist-react/pull/1680/files#r381927603 - where we are reacting at the keystroke level to issue a query back to a remote API. One second seemed long at first, but I think it's correct given how fast people might/might not type.

client-app/src/examples/portfolio/PortfolioPanelModel.js
```diff
@@ -62,7 +62,7 @@ export class PortfolioPanelModel {
             run: (position) => {
                 this.detailPanelModel.setPositionId(position ? position.id : null);
             },
-            delay: 500
+            debounce: 500
```
amcclain: This will be a debounce on top of the grid's built-in selection debounce. Do we really need it now that the grid debounce will be working as intended? Maybe loading all the position details is intensive enough that it's warranted still?

lbwexler: To be clear, the grid has just a hyper minimal debouncing (0ms) between the agGrid and our selection model, just to bulk up multi-selection events that are generated by agGrid as streams of events.

This much larger debouncing here, is really aimed at keeping the master detail from going crazy with network requests.

lbwexler: reducing it to 300 per our early convo -- improved debounce behavior does warrant that!

PR #369, opened 2020-02-26
Update toolbars to work with overflow.

TomTirapani:
Issue: https://github.com/xh/hoist-react/issues/1524
Hoist-React PR: https://github.com/xh/hoist-react/pull/1719


lbwexler:
Tom and I discussed this, and we are inclined to hold off on this until v.31.  We also discussed potentially making it opt-in

PR #384, opened 2020-03-20
Revert "Update to PanelModel api"

lbwexler:
Reverts xh/toolbox#382


lbwexler:
Turns out the Hoist Change was not quite ready!  Finding some issues in the testing.
Reverting

PR #411, opened 2020-06-11
Add demo of autoGroupColumnDef with groupUseEntireRow: false to SampleGrid

cnrudd:
As discussed, added autoGroupColumnDef and groupUseEntireRow switch.

Am wondering if we might also add a switch for groupMultiAutoColumn: true, to demo/test this feature: https://www.ag-grid.com/javascript-grid-grouping/#multi-auto-column-group ?


lbwexler:
Looks good.  Let's discuss this pattern of how we get the grid to re-render when we change agOptions.  Its something applications will want to do, and this can be a great example of that as well.  Want to make it as clean as possible...

PR #438, opened 2020-10-22
Search from nav with options in SoftConfig

cnrudd:
This branch puts a search bar in the top nav - right corner.
The options presented by the search box are set up in soft config.
It's a nearly 400 line JSON soft config entry that includes all the input components, date & number formats, tab panel tab variations, etc.    Very handy to have in  soft config.

The soft config entry is already in https://toolbox-dev.xh.io/admin/config under Toolbox => searchOptions if you want to have a look.

It might be interesting to improve the experience by adding an options renderer that styles components vs props vs methods to help users identify their results.


lbwexler:
Colin and I agreed the ship has sailed on this one.  Replaced by the new doc viewer

PR #446, opened 2020-11-24
Use GroupingChooser component

TomTirapani:
See hoist-react PR: https://github.com/xh/hoist-react/pull/2200


lbwexler:
Code and UI look great to me.  Some thoughts, in no particular order, which I would like to discuss in person

+ I sort of do miss the indenting a bit.  Wonder if it worth bringing that back.  Could also help signal where the new value is going to go.
+ looking for some kind of tiny "ok/cancel" buttons (can be just icons).  I know others may not agree.  Just really nice after investing the time to set something up to click on a surface that I know will safely save it, rather than being uncertain about clicking somewhere else.  Also -- especially because we don't actually re-render the inline value immediately, I really don't have confidence that my change has been accepted.  
+ One nice side effect of above, is that if I put things in to a bad state (e.g. remove all dimensions), we could indicate that to the user by *not* having the ok checkbox enabled.
+ Could consider leaving the add visible, but disabled with appropriate message when adding not possible (max Depth, leaf dimension already chosen, etc)

PR #450, opened 2020-11-26
Mobx6 and remove decorators

lbwexler:


TomTirapani:
Merging this to keep toolbox develop compatible with hoist-react develop

PR #460, opened 2020-12-30
Add 'group' to filterable fields in leftRightChooserPanel example

petradish:


lbwexler:
Seems like a correct change -- although not sure why Hoist makes the user specify these fields, rather than just defaulting them.

PR #465, opened 2021-02-11
Allow variable height in grid rows.

cnrudd:
This is a WIP for this Hoist-Reactd issue: https://github.com/xh/hoist-react/issues/2196

This branch adds some hand written comments to the standard grid: http://localhost:3000/app/grids/standard

and some auto generated comments to the ag-grid:
http://localhost:3000/app/grids/agGrid

So that variable height rows can be demoed/tested.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -10,7 +10,7 @@ within, does not offer public support or a license at this time, makes no warran
 as to the functionality, correctness, or fitness of this software, and shall not be held liable for
 its use outside the scope of a structured consulting engagement.
 
-Copyright © 2020 Extremely Heavy Industries Inc. - all rights reserved
+Copyright © 2021 Extremely Heavy Industries Inc. - all rights reserved
 
  ------------------------------------------
 
--- a/README.md
+++ b/README.md
@@ -24,4 +24,4 @@ or another OAuth based solution.
                                                                                   
 📫☎️🌎 info@xh.io | <https://xh.io/contact>
 
-Copyright © 2020 Extremely Heavy Industries Inc.
+Copyright © 2021 Extremely Heavy Industries Inc.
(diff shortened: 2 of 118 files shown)


amcclain:
Hi - just started to review the autoheight PR and fired up Toolbox on the branch to see it in action. 

Our little sample grid component is designed to provide the best introduction to our use of the grid and its core features, to make a great impression all around for someone browsing Toolbox to learn about what we do and all our best practices, and be nicely re-usable in a few contexts where we want to demo a grid.  

Upon starting it up, I now see:
Screen Shot 2021-02-19 at 1 38 33 PM

Here it is in our docked view example:
Screen Shot 2021-02-19 at 1 48 58 PM

I understand and appreciate the need to demo the auto-height support, but this kind of thing stops me completely dead in my tracks when it comes to reviewing and merging these updates. We really need to take the full picture into account with Toolbox and the whole Hoist project in general. 

Let me look a bit more at the change and I'm happy to suggest some options for how we might better demo/test it. What comes to mind is dedicated tab for testing auto-heights with a much simpler example where longer text content is generated locally. All of the positionService / rawPosition stuff on the server is primarily about being able to generate large / realistic-ish data for testing things like streaming cube updates and aggregations. 

So again - let's connect again on it next week, but I wanted to provide this immediate feedback so you can see where I am coming from when I pick these things up, and the kinds of things that can 💥  my mind.  😃

TomTirapani:
I feel this is partially my fault - I didn't realise the Toolbox branch was intended to be merged, thinking it was a local testbed for a client app use case. Otherwise I would have taken a closer look at the code and user experience here.

As I said in the issue https://github.com/xh/hoist-react/issues/2196, I don't think we should advertise our support for auto height. It's currently enabled via passing two agOptions to a column. If we want to make it a "first class" part of the Toolbox, it would need to become a supported part of the API and at that point we should definitely add a Toolbox example.

PR #474, opened 2021-04-01
Add shared AppBanner

TomTirapani:
See hoist-react PR: https://github.com/xh/hoist-react/pull/2386


amcclain (approved):
This looks great, thanks for updating - will merge when we merge the hoist-react PR.

PR #481, opened 2021-05-19
Contact example app

dsilberger:
New Example App: Contact

- Add new example app to toolbox


client-app/src/examples/contact/TileView.js
```diff
@@ -0,0 +1,39 @@
...
+
+        return tileFrame({
+            spacing: 10,
+            desiredRatio: 1.25,
+            minTileWidth: 300,
+            maxTileWidth: 600,
+            minTileHeight: 250,
+            maxTileHeight: 250,
+            items: store.records.map(record => {
+                const isSelected = selModel.selectedRecordId === record.id;
```
amcclain, line 18: No difference in functionality, but you could destructure `{selectedRecord}` off of the model at the top and then compare each record to it directly.

client-app/src/examples/contact/TileView.js
```diff
@@ -0,0 +1,39 @@
...
+                const {profilePicture, name} = record.data;
+
+                return vframe({
+                    style: {backgroundImage: `url(${profilePicture})`},
```
amcclain, line 22: Anything go wrong here if `!profilePicture`?

client-app/src/examples/contact/TileView.js
```diff
@@ -0,0 +1,39 @@
...
+                    alignItems: 'center',
+                    justifyContent: 'center',
```
amcclain, line 24: Since we are already applying a CSS class to this element, I would push any rules like this into the stylesheet.

client-app/src/examples/contact/TileView.scss
```diff
@@ -0,0 +1,19 @@
...
+    border-color: var(--xh-orange);
+  }
+
+  .floating-name {
+    position: absolute;
+    font-weight: bold;
+    text-shadow: 2px 0 0 #000, 0 -2px 0 #000, 0 2px 0 #000, -2px 0 0 #000;
+    right: 20px;
+    bottom: 10px;
+  }
```
amcclain, line 18: These were looking weird when I ran locally:

Screen Shot 2021-05-19 at 4 20 48 PM

Here's a quick pass at that approach I mentioned using a semi-transparent background on the name div - could still be tuned a bit I am sure, but looks promising and could also hold a star to indicate / toggle fave status.

Screen Shot 2021-05-19 at 4 27 41 PM

client-app/src/examples/contact/detail/DetailsPanel.js
```diff
@@ -0,0 +1,116 @@
...
+            icon: Icon.detail(),
+            width: 325,
+            flex: 'none',
+            items: currentRecord ? renameThisProfile({record: currentRecord}) : placeholder('Select a contact to view their details.'),
+            ...props
+        });
+    }
+});
+
+const renameThisProfile = hoistCmp.factory({
```
amcclain, line 31: Time to... renameThisProfile 😄

client-app/src/examples/contact/detail/DetailsPanel.js
```diff
@@ -0,0 +1,116 @@
...
+    model: uses(DetailsPanelModel),
+    className: 'contact-details-panel',
+
+    render({model, className, ...props}) {
+        const {currentRecord} = model;
+
+        return panel({
+            title: currentRecord?.data.name ?? 'Select a contact',
+            className,
+            icon: Icon.detail(),
```
amcclain, line 22: This little misalignment is a bit unfortunate here - 

Screen Shot 2021-05-19 at 4 32 46 PM

I'm thinking we might not need / want the title at all - it doesn't add much and I think looks better (slightly) without. 

Alternative would be to force its height to line up with the toolbar - I am a bit sad that's not the case out of the box. Or to try with the detail panel below the grid in a more horizontal layout - we might want that if you add that tags column, so we have more width for the grid.

client-app/src/examples/contact/detail/DetailsPanel.js
```diff
@@ -0,0 +1,116 @@
...
+                        }),
+                        formField({
+                            field: 'homePhone',
+                            readonlyRenderer,
+                            item: textInput()
+                        }),
+                        formField({
+                            field: 'bio',
+                            readonlyRenderer,
+                            item: textArea()
```
amcclain, line 87: Whenever I have a long series of configs like this with a fair amount of duplication, I look to DRY up by spreading a partial with the defaults - e.g.

```
const ffConf = {
  readonlyRenderer: (val) => val ?? '-', 
  item: textInput()
};
```

Then you can really tighten up these formFields:

```
items: [
    formField({...ffConf, field: 'name'}),
    formField({...ffConf, field: 'email'}),
    formField({...ffConf, field: 'location'}),
    formField({...ffConf, field: 'workPhone'}),
    formField({...ffConf, field: 'cellPhone'}),
    formField({...ffConf, field: 'homePhone'}),
    formField({...ffConf, field: 'bio', item: textArea()})
]
```

client-app/src/examples/contact/detail/DetailsPanel.scss
```diff
@@ -0,0 +1,24 @@
+.contact-details-panel {
+  &__inner {
+    height: 100%;
+    padding: var(--xh-pad-double-px);
+    overflow-y: auto;
+  }
+
+  // make this centered again
```
amcclain, line 8: I also noticed that it will trigger scrolling if wider than the detail panel width.

client-app/src/examples/contact/detail/DetailsPanelModel.js
```diff
@@ -0,0 +1,70 @@
+import {XH, HoistModel, managed} from '@xh/hoist/core';
+import {makeObservable, observable, action} from '@xh/hoist/mobx';
+import {FormModel} from '@xh/hoist/cmp/form';
+
+export class DetailsPanelModel extends HoistModel {
+
+    @observable
+    profilePicture
```
amcclain, line 8: Dead code, right?

client-app/yarn.lock
```diff
@@ -240,25 +240,25 @@
   dependencies:
     "@babel/highlight" "^7.12.13"
 
-"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.13.15", "@babel/compat-data@^7.13.8", "@babel/compat-data@^7.14.0":
```
amcclain, line 243: This change to `yarn.lock` is the result of you running `yarn upgrade` locally, which is totally fine to do, but ideally not part of this diff (assuming that the upgraded deps are not important to the new functionality). 

I'd revert this file on the branch as a best practice.

If you did want to ensure that Toolbox was set to use the latest Hoist, you can run `yarn upgrade` on the develop branch, verify and push that, then do a catch-up merge to get the contacts branch even with develop.

Of course this would *not* apply if the PR included a deliberate change to dependencies that should get merged along with it.

grails-app/controllers/io/xh/toolbox/contacts/ContactsController.groovy
```diff
@@ -0,0 +1,40 @@
...
+@Access(['APP_READER'])
+class ContactsController extends BaseController {
+
+    def configService
+
+    def index() {
+        renderJSON(getContacts())
+    }
+
+    def update(String id) {
```
amcclain, line 20: Let's add a comment to this method to cover this rather unusual use of configs (which I know I introduced!)

```suggestion
    /**
     * Updates the requested contact with any new values provided in the request body.
     * 
     * NOTE - this is a very unusual use of the Hoist config system and would NOT be recommended 
     * a production application. Configs are typically only modified interactive by users via the
     * Hoist admin console. Here we are storing example contact records in a config for convenience 
     * (mainly to avoid setting up GORM objects and their DB tables). This code patches that config 
     * in place, allowing the Contacts example app to demo data updates.
     */
    def update(String id) {
```

grails-app/controllers/io/xh/toolbox/contacts/ContactsController.groovy
```diff
@@ -0,0 +1,40 @@
...
+        if (!user.isHoistAdmin) throw new NotAuthorizedException()
```
amcclain, line 21: @lbwexler do we support `@Access` annotations at the method level?  I assumed not - is this the best practice approach if we want to avoid creating an entirely different controller?

amcclain (commented):
Looking great!  

A few comments below with areas that I think could be improved or would be generally interesting to explore.  Looking forward to seeing the next iteration with the tags.

grails-app/controllers/io/xh/toolbox/contacts/ContactsController.groovy
lbwexler: yep, you got it. So I think this is what we want -- yes!

lbwexler:
Looking at this now

lbwexler:
Reviewed with Dan and we are ready to Merge.  

This will get this on to to develop, and allow us fill it out with proper data for the entire team.  That will hopefully inform any final tweaks we want to do before release.

PR #482, opened 2021-05-26
Demo New comma options in number formatters

febbraiod:
Depends on https://github.com/xh/hoist-react/pull/2463


febbraiod:
Diff looks ok, but if anyone knows why my commit history looks so whacked I'd like to hear it

PR #484, opened 2021-05-28
Todo app

lbwexler:


client-app/src/examples/todo/TaskDialogModel.js
```diff
@@ -0,0 +1,103 @@
...
+    async submitAsync() {
+        const {formModel, parentModel} = this,
+            {values} = formModel,
+            isValid = await formModel.validateAsync();
+
+        if (isValid) {
+            const {description, dueDate, complete} = values,
+                existingId = values.id,
+                task = {
+                    id: existingId ?? Date.now(),
```
lbwexler, line 60: what is our standard here for generating unique ids on client -- this was my quick idea.

client-app/src/examples/todo/TaskDialogModel.js
```diff
@@ -0,0 +1,103 @@
...
+                    description,
+                    dueDate,
+                    complete: complete ?? false
+                };
+
+            existingId ? await parentModel.editTaskAsync(task) : await parentModel.addTaskAsync(task);
```
lbwexler, line 66: better to use if else here.  better to use ternary when we are actually using the value.

client-app/src/examples/todo/TodoPanel.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const todoPanel = hoistCmp.factory({
+    model: creates(TodoPanelModel),
+
+    render({model}) {
+        return fragment(
+            taskDialog(),
+            panel({
+                tbar: tbar(),
+                item: grid({onRowDoubleClicked: (e) => model.taskDialogModel.openEditForm(e.data.data)}),
```
lbwexler, line 20: I like this form of openEditForm() better -- where we pass the task, rather than get from selection.  But need to actually change the implementation, which is ignoring this argument and getting from selection.  Note that we need to also account for the fact that group rows can be clicked and nothing should happen in that case.

client-app/src/examples/todo/TodoPanel.js
```diff
@@ -0,0 +1,72 @@
...
+                mask: 'onLoad',
+                bbar: bbar()
+            })
+        );
+    }
+});
+
+const tbar = hoistCmp.factory(
+    /** @param {TodoPanelModel} */
```
lbwexler, line 29: don't think we want this docstring here.

client-app/src/examples/todo/TodoPanelModel.js
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    async removeSelectedTasksAsync() {
+        const {selectedTasks} = this;
+
+        if (isEmpty(selectedTasks)) return;
+
+        const count = selectedTasks.length,
+            {description} = selectedTasks[0],
+            message = count === 1 ? `'${description}?'` : `${count} tasks?`,
```
lbwexler, line 147: might want to display only certain number of chars of description

client-app/src/examples/todo/TodoPanelModel.js
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    info(message) {
+        XH.toast({message});
+    }
+
+    dueDateRenderer(v, {record}) {
+        const overdue = v && v < LocalDate.today() && !record.data.complete,
+            dateStr = fmtCompactDate(v?.date, {
+                sameDayFmt: '[Today]',
```
lbwexler, line 255: meant to say in our review -- if we are going to be explicit about specifying custom `sameDayFmt` and `distantFmt`, we should also just explicitly specify `nearFmt`.

client-app/src/examples/todo/TodoPanelModel.js
```diff
@@ -0,0 +1,268 @@
...
+            {description} = selectedTasks[0],
+            message = count === 1 ? `'${description}?'` : `${count} tasks?`,
+            confirmed = await XH.confirm({
+                title: 'Confirm',
+                message: `Are you sure you want to permanently remove ${message}`
+            });
+
+        if (!confirmed) return;
+
+        const label = count === 1 ? `Task removed: '${description}'` : `${count} tasks removed`;
```
lbwexler, line 155: can use a utility method for `labelForTask` to get the (potentially) shortened description, that was also mentioned above.

client-app/src/examples/todo/TodoPanelModel.js
```diff
@@ -0,0 +1,268 @@
...
+
+        for (const task of selectedTasks) {
+            await XH.todoService.removeTasksAsync(task);
```
lbwexler, line 158: pluralization in this method name is wrong, as it currently only does one - but given how we are actually using it here, would make sense to actually go ahead and implement the bulk delete at the service 

In a real world system that would be a lot safer -- want to modify state/backend in one shot whenever it is easier to do so -- we can discuss

client-app/src/desktop/tabs/examples/ExamplesTab.js
```diff
@@ -102,6 +102,19 @@ function getExamples() {
...
                 </p>
             ]
+        },
+        {
+            title: 'Todo List',
+            icon: Icon.clipboard(),
+            path: '/todo',
+            text: [
+                <p>
+                    This example provides a simple to-do application. Provides examples of [replace later...filtering and searching data from an external API.]
```
lbwexler, line 112: cut/paste mistake with text here

client-app/src/examples/todo/TaskDialogModel.js
```diff
@@ -0,0 +1,103 @@
...
+        fields: [
+            {
+                name: 'id'
+            },
+            {
+                name: 'complete'
+            },
+            {
+                name: 'description',
+                rules: [required, lengthIs({max: 1000})]
```
lbwexler, line 31: could also just make this length much shorter, for purposes of the demo app

client-app/src/examples/todo/TodoService.js
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Service to manage fetching, updating and persisting tasks.
+ * Tasks are persisted for each user using the Hoist preference system.
+ */
+export class TodoService extends HoistService {
+
+    /** @returns {Promise<TaskItem[]>} */
+    async getTasksAsync() {
+        return XH.getPref('todoApp').map(it => {
+            const dueDate = isString(it.dueDate) ? LocalDate.get(it.dueDate) : null,
```
lbwexler, line 15: We should just make LocalDate.get() nil safe.  There must be a lot of code like this around.  I will file a ticket.

PR #492, opened 2021-06-14
Contacts update

lbwexler:


lbwexler:
Reviewed these changes with Dan -- mostly clarifying refactors, breaking down components etc.

PR #529, opened 2021-09-09
Fix breaking change on autosizeMaxWidth

petradish:
See https://github.com/xh/hoist-react/pull/2660


petradish:
Closing this due to API change

PR #532, opened 2021-09-21
Add new color tree style options

petradish:
Adds `TreeStyle.COLORS` and `TreeStyle.COLORS_AND_BORDERS` styles to grid options
See Hoist-React PR: https://github.com/xh/hoist-react/pull/2653


amcclain (commented):
Eager to get this merged!

+ I am running the HR and TB branches and don't see the group borders at all when selecting those options - are you?  (I don't see how this change would have broken those, but they are working on toolbox-dev.)
+ To really appreciate this change, we need at least one (or ideally two) levels of additional grouping in our Toolbox example. Could you please look to see if you can come up with something? I'd suggest Long/Short and Profit/Loss - two easy dimensions based on the numeric data that doesn't require faking anything new.

edit - meant this for the HR PR - will add there now.......

petradish:
Added a couple more string dimensions to the sample grid grouping chooser. Added 2 available properties - fund and trader - from RawPosition instead of creating anything new there.

PR #533, opened 2021-09-22
Grails5

lbwexler:


build.gradle
```diff
@@ -22,85 +21,131 @@ group 'io.xh'
 apply plugin:'idea'
 apply plugin:'war'
 apply plugin:'org.grails.grails-web'
-apply plugin:'org.grails.grails-gsp'
 
 repositories {
     mavenLocal()
-    mavenCentral()
```
amcclain, line 29: Why are we removing mavenCentral here and above?  Do we want to ensure we're getting everything from the Grails repo (assuming it has a mirror of all the deps we need)?

For the record, I would not want to lean any more on our XH repo as a pass-through source for dependencies - hopefully we can do away with that soon altogether.

build.gradle
```diff
@@ -22,85 +21,131 @@ group 'io.xh'
...
     maven {url 'https://repo.grails.org/grails/core'}
     maven {url 'https://repo.xh.io/content/groups/public/'}
 }
 
-configurations.all {
-    resolutionStrategy.cacheChangingModulesFor 0, 'seconds'
```
amcclain, line 35: This gave us the behavior of looking for new snapshot versions each time dependencies were resolved.
Looks like the default [remains 24 hours](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.ResolutionStrategy.html#org.gradle.api.artifacts.ResolutionStrategy:cacheChangingModulesFor(int,%20java.util.concurrent.TimeUnit)), unless there is some other way this is being set, which might be unexpectedly long when using any snapshot builds.

build.gradle
```diff
@@ -22,85 +21,131 @@ group 'io.xh'
...
-    // Include Hoist as an inline plugin in multi-project mode when running in development
-    grails {
-        plugins {
-            compile project(":hoist-core")
+
+// Tasks to create distribution bundles that include built client-assets, as well as (for the
+// first task) the WAR. This creates a self-contained archive with both sides of the build output.
+// NOTE - these rely on the client-side build having been completed (outside of Gradle) by Yarn and
+// Webpack. A future project could be to orchestrate the Webpack build from within Gradle as well.
+apply plugin:'distribution'
```
amcclain, line 72: I would be happy to remove all of this complexity around publishing from the Gradle file. We added this as part of research for a client who wanted some custom artifact publishing for app builds, but I am not sure they are actually using it, and either way I don't think it's standard enough for us to bother including in Toolbox as an example.

Happy to let the Docker container images be our versioned outputs from the builds.

gradle.properties
```diff
@@ -2,10 +2,14 @@ xhAppCode=toolbox
...
-gormVersion=6.1.11.RELEASE
-gradleWrapperVersion=4.10.3
+grailsVersion=5.0.3
+groovyVersion=3.0.7
+gormVersion=7.1.0
+grailsGradlePluginVersion=5.0.0
+
+org.gradle.daemon=true
+org.gradle.parallel=true
+org.gradle.jvmargs=-Dfile.encoding=UTF-8 -Xmx1024M
```
amcclain, line 15: Just checking that we are sure this value makes it through OK without any extra quoting (the inner = seems like it could cause problems reading it here on GH)

build.gradle
lbwexler: will restore this -- great catch.

gradle.properties
lbwexler: will look -- great catch -- this was copied from output from grails create-app, but good to check

PR #547, opened 2022-01-12
MySQL driver and datasource properties updated to v8

amcclain:
+ Upgraded MySQL instances have better defaults re. character encoding (`utf8mb4` for proper unicode/emoji support), and in general we were due for an upgrade here.
+ Remove defaulted properties from JDBC URL, move `useSSL` flag to driver properties block along with the others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## v3.0-SNAPSHOT - unreleased
 
+### Technical
+* MySQL driver and datasource properties updated to MySQL v8.
+
 ## v2.13.0 - 2021-01-11
 
 ### New Features
--- a/build.gradle
+++ b/build.gradle
@@ -58,8 +58,8 @@ if (inlineHoist) {
 }
 
 dependencies {
-    implementation "org.bitbucket.b_c:jose4j:0.7.2"
-    runtimeOnly "mysql:mysql-connector-java:5.1.47"
+    implementation "org.bitbucket.b_c:jose4j:0.7.9"
+    runtimeOnly "mysql:mysql-connector-java:8.0.27"
 
     // Support for reloading app on file change in dev environment
     developmentOnly(
--- a/grails-app/conf/runtime.groovy
+++ b/grails-app/conf/runtime.groovy
@@ -83,8 +83,8 @@ dataSource {
         //------------------------
         // Remove connections considered to be abandoned...
         removeAbandoned = true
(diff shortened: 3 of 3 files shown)


amcclain:
To get my local DB environment back in shape, I did the following.
+ Uninstalled the older v5.6 install I had previously.
  + I had installed via homebrew, so uninstalled via `brew uninstall mysql@5.6`
  + Homebrew uses `/usr/local/var/mysql` for the data/conf directory. To start completely fresh, I deleted this directory on my machine after uninstalling. (⚠️ NOTE - this will delete any local databases/data you have - pls consider carefully if you have other projects.)
+ Reinstalled via homebrew with `brew install mysql`
  + Then start with `brew services start mysql`
  + Then secured (optional) to set root pwd, etc with `mysql_secure_installation`
+ Created empty toolbox db.
  + `mysql -u root -p` (then enter pwd set above, if any)
  + `create database toolbox;` (v8 will use a proper charset by default)
  + `quit;`
+ If you are OK setting up your `toolbox.yml` file to connect with the root user (fine for local dev), then you should be set - root will have the rights to connect to the DB and create the schema. If you want a dedicated DB user for the app you'll need to create and assign dbo on the new DB.
+ Run the app off of this branch to let it recreate the schema fresh with the latest driver / Hibernate / dialect. Then stop the app.
+ Import the existing data. I snapped an export from "prod" Toolbox this AM that has the data only - i.e. no schema create/drop. That's saved in https://drive.google.com/drive/folders/1aZ9Y5aSF-ehsYQJF0CnmHzj1rpRD-bFD?usp=sharing. If you e.g. copy that file to someplace convenient like your desktop you can then run:
  + `mysql -u root -p toolbox < ~/Desktop/toolbox-data-only_20220112_0939PST.sql`
+ Restart the app, test your new setup by editing e.g. an app config value - be sure to try an 😀!

amcclain:
I have created a new MySQL v8 DB within our AWS account. When we are ready to merge this and cutover, I'm happy to repoint our container configs there to point to the new DB and ensure we have it loaded with a backup as described above.

cnrudd:
On my dev machine I've been using MariaDB v10.5.6 installed from Homebrew since October 2020.  A look through all of the tables in my local toolbox database shows that all text columns are using character_set utf8mb4.   I pulled and ran this mysql8 branch and it is running well.  I am able to put emojis into configs.

TomTirapani:
Got my local toolbox running against v8 without too much trouble!

PR #548, opened 2022-01-13
Improvements to dev-time support

lbwexler:
More verifiably enable/disable Spring developer tools

+ https://stackoverflow.com/questions/59252470/gradle-spring-boot-devtools-developmentonly-and-runtimeclasspath

### DIFF
--- a/grails-app/conf/runtime.groovy
+++ b/grails-app/conf/runtime.groovy
@@ -18,7 +18,7 @@ def dbHost = getInstanceConfig('dbHost') ?: 'localhost',
     dbSchema = getInstanceConfig('dbSchema') ?: appCode,
     dbUser = getInstanceConfig('dbUser') ?: appCode,
     dbPassword = getInstanceConfig('dbPassword'),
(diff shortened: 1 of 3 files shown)


grails-app/conf/runtime.groovy
```diff
@@ -18,7 +18,7 @@ def dbHost = getInstanceConfig('dbHost') ?: 'localhost',
     dbSchema = getInstanceConfig('dbSchema') ?: appCode,
     dbUser = getInstanceConfig('dbUser') ?: appCode,
     dbPassword = getInstanceConfig('dbPassword'),
-    sslSuffix = dbHost == 'localhost' ? '&useSSL=false' :  ''
+    sslSuffix = dbHost == 'localhost' ? '&useSSL=false&allowPublicKeyRetrieval=true' :  ''
```
amcclain, line 21: This bit has been improved in the mysql8 branch, where we consistently spec these driver-specific properties in the config block below (vs. URL query params).  Will merge that now and then can update/merge this PR.

PR #549, opened 2022-01-17
Use new API for models

lbwexler:
+ Remove Repeated State in portfolio app.
+ Lazy create/decouple read-only sub-view


client-app/src/admin/tests/asyncLoops/AsyncLoopPanel.js
```diff
@@ -20,7 +20,7 @@ import {SECONDS} from '@xh/hoist/utils/datetime';
  * Test to demonstrate use of {@see whileAsync} and its more common cousin {@see forEachAsync}.
  */
 export const asyncLoopPanel = hoistCmp.factory({
-    model: creates(() => new Model()),
+    model: creates(() => AsyncLoopPanelModel),
```
febbraiod, line 23: Any reason this can't be:

`creates(AsyncLoopPanelModel)`

client-app/src/admin/tests/dataview/DataViewTestPanel.js
```diff
@@ -7,7 +7,7 @@ import {slider} from '@xh/hoist/desktop/cmp/input';
 import './DataViewTestPanel.scss';
 
 export const dataViewTestPanel = hoistCmp.factory({
-    model: creates(() => new Model()),
+    model: creates(() => DataViewTestPanelModel),
```
febbraiod, line 10: see above

client-app/src/desktop/common/Wrapper.js
```diff
@@ -13,12 +13,10 @@ import './Wrapper.scss';
 export const [Wrapper, wrapper] = hoistCmp.withFactory({
     displayName: 'Wrapper',
     className: 'tbox-wrapper xh-tiled-bg',
-    model: false,
-
-    render({className, description, links, children, ...rest}) {
-
-        const localModel = useLocalModel(() => new Model(links));
+    render({className, description, children, ...props}) {
+        const impl = useLocalModel(Model);
```
febbraiod, line 17: Should this be "WrapperModel". Also why does this one need to use "useLocalModel"

client-app/src/desktop/tabs/other/AppNotificationsPanel.js
```diff
@@ -77,7 +77,7 @@ const idleServiceTile = hoistCmp.factory(
                     className: 'tb-idle__button',
                     icon: Icon.userClock(),
                     text: 'Enter Sleep Mode',
-                    onClick: () => XH.idleService.suspendApp()
+                    onClick: () => XH.suspendApp({reason: 'IDLE'})
```
febbraiod, line 80: unrelated to model change yeah?

client-app/src/examples/portfolio/detail/DetailPanelModel.js
```diff
@@ -1,17 +1,16 @@
...
 
 export class DetailPanelModel extends HoistModel {
 
-    @bindable positionId = null;
+    @observable positionId = null;
 
-    @managed ordersPanelModel = new OrdersPanelModel();
-    @managed chartsPanelModel = new ChartsPanelModel();
+    @lookup(PortfolioPanelModel) parentModel;
+    @managed ordersPanelModel = new OrdersPanelModel(this);
```
febbraiod, line 13: This looks like the anti-pattern to the change here. Aren't we just create this here so that it can have a reference to the parent? Shouldn't OrdersPanelModel just lookUp the DetailPanelModel?

febbraiod: Maybe anti-pattern isn't right, but rather "the old way"

client-app/src/examples/portfolio/detail/DetailPanelModel.js
```diff
@@ -1,17 +1,16 @@
...
+import {PortfolioPanelModel} from '../PortfolioPanelModel';
 
 export class DetailPanelModel extends HoistModel {
 
-    @bindable positionId = null;
+    @observable positionId = null;
 
-    @managed ordersPanelModel = new OrdersPanelModel();
-    @managed chartsPanelModel = new ChartsPanelModel();
+    @lookup(PortfolioPanelModel) parentModel;
```
febbraiod, line 12: Did we talk about not calling this "parentModel" anymore but rather naming it base don its class name? I actually don't mind parentModel, but it seems like not really the case anymore since DetailPanelModel doesn't get instantiated by the "parent" anymore.

client-app/src/examples/portfolio/detail/OrdersPanelModel.js
```diff
@@ -17,14 +16,18 @@ import {
...
     @managed gridModel;
     @managed filterChooserModel;
 
-    @bindable positionId = null;
+    get positionId() {
+        return this.parentModel.positionId;
+    }
 
-    constructor() {
+    constructor(parentModel) {
```
febbraiod, line 27: see above

client-app/src/examples/portfolio/detail/charts/LineChart.js
```diff
@@ -1,15 +1,27 @@
...
+            item: chart(),
+            mask: 'onLoad',
+            flex: 1
+        });
+    }
+});
+
+class LineChartModel extends HoistModel {
+
+    @lookup(ChartsPanelModel) parentModel;
```
febbraiod, line 20: see above.

client-app/src/mobile/popover/PopoverPage.js
```diff
@@ -71,16 +70,11 @@ const popoverContent = hoistCmp.factory({
 /**
  * LocalModel used to demonstrate opening a Popover in controlled mode
  */
-class LocalModel extends HoistModel {
-    @observable isOpen = false;
+class PopoverPageModel extends HoistModel {
+    @bindable isOpen = false;
```
febbraiod, line 74: wondering about this choice. Thought the idea was to only used bindables for cases where we use the hoist "bind" magic.

client-app/src/admin/tests/asyncLoops/AsyncLoopPanel.js
lbwexler: Yes, need closure, because not defined until further in this file.

client-app/src/admin/tests/dataview/DataViewTestPanel.js
lbwexler: see above

client-app/src/desktop/common/Wrapper.js
lbwexler: yes, and yes.  good suggestions

client-app/src/desktop/tabs/other/AppNotificationsPanel.js
lbwexler: yep -- probably should have fixed directly against develop

client-app/src/examples/portfolio/detail/DetailPanelModel.js
lbwexler: I don't think parentModel is now a verboten term. Let's stick with it to avoid churn

client-app/src/examples/portfolio/detail/DetailPanelModel.js
lbwexler: this was an example where parent actually needed a reference to the OrdersPanelModel.  could have done it in a number of ways, chose to leave this one like this.

client-app/src/mobile/popover/PopoverPage.js
lbwexler: we are using it on line 36 above from the popover/

PR #554, opened 2022-01-31
Fix default groupBy in Grouped Columns Grid and ActivityWidget

TomTirapani:
See issue: https://github.com/xh/hoist-react/issues/2872


lbwexler:
Nice Tom!!!  Beautiful example of the value of minimizing state as well.

PR #561, opened 2022-03-01
Revised Examples Tab

amcclain:


amcclain:
Hi @dsilberger - I pulled this today to have a look - note that I did a simple catch-up merge with develop, so please be sure to pull.

Let's plan to look at this together tomorrow. I think with a few quick updates we'll be in a good place to merge. Top things I would recommend:

* Let's try it with the list of examples docked to the left and given the full height of the overall tab. The iFrame itself can then be inset with a chunky margin around it, equal on all sides - something like 30px maybe.
* As per the TODO in the code, quick refactor to break out an app tile component, then some styling to set off the title and icon.
* We could consider always showing the full version of the tile - so you can scan and see the (brief) descriptions for each without needing to click each one.  Clicking activates the tile, and maybe we still only show the "launch" button when the tile is active, so that appears and only then gives you the option to open in a new (browser) tab.

All things we can discuss together, of course.

client-app/src/desktop/tabs/examples/ExamplesTab.js
```diff
@@ -1,138 +1,230 @@
...
+
 
 export const examplesTab = hoistCmp.factory(
-    () => wrapper(
-        hbox({
-            className: 'example-tile-container',
-            flexWrap: 'wrap',
-            items: getExamples().map(e => panel({
+    () => {
+        const impl = useLocalModel(LocalModel);
```
lbwexler, line 21: Make this LocalModel a named model "ExampleTabModel" and promote it to be the actual model for the tab.  This is our new best practice and will allow you to avoid having to pass it explicitly to your sideBar and appTileBar

client-app/src/desktop/tabs/examples/ExamplesTab.js
```diff
@@ -1,138 +1,230 @@
...
+        })),
+        button({
+            icon: Icon.chevronRight(),
+            onClick: () => model.leftPanelModel.collapsed = false
+        })],
+        vertical: true
+    });
+});
+
+class LocalModel extends HoistModel {
```
lbwexler, line 98: And this model should go in its own file --   Its complex and important enough -- not an implementation detail for the component.

client-app/src/desktop/tabs/examples/ExamplesTab.js
```diff
@@ -1,138 +1,230 @@
...
-    ];
+    )
+
+    get examples() {
+        return [
+            {
+                title: 'Portfolio',
+                icon: Icon.portfolio(),
+                path: '/portfolio',
+                cmp: () => portfolioPanel(),
```
lbwexler, line 118: this cmp argument is cruft?  We aren't doing this anymore -- the main point of using our iFrame

client-app/src/examples/contact/details/DetailsPanel.js
```diff
@@ -11,7 +11,7 @@ import {isEmpty} from 'lodash';
 import './DetailsPanel.scss';
 import {DetailsPanelModel} from './DetailsPanelModel';
 
-export const detailsPanel = hoistCmp.factory({
+    export const detailsPanel = hoistCmp.factory({
```
lbwexler, line 14: looks like some stray whitespace

lbwexler:
Dan -- made some comments.  I am sure we will tweak this further before release,  but am eager to get it merged

PR #579, opened 2022-04-25
Added an empty tabContainer into mobile container page

jskupsik:


lbwexler:
Hi Jakub -- 

Am going to close this PR -- think that anything that we actually merge to toolbox needs to be a little bit more obvious and userful to the end user of toolbox that is trying to learn the toolkit/api.

One other important thing -- you *never* want to instantiate an actual model (with new)  in the render method of a component.  In practice, that will mean that the model is recreated on every render.

If you would like to create a model lazily that will be maintained in react state, you can specify a config for a model (i.e.  the arguments to its constructor), e.g. `model: {tabs: []}`

PR #580, opened 2022-04-27
Replace Popover menus with new MenuButton component

ghsolomon:


lbwexler:
Closing — stale since 2022. Targets `.js` files that have since been migrated to TypeScript, conflicts with `develop`, and depends on a generic `MenuButton` component that never landed in hoist-react. The Popover→menu cleanup it proposed can be redone fresh against the current `.ts` files if still desired.

PR #582, opened 2022-05-05
Update gails artifactory URL in build.gradle

ghsolomon:


amcclain:
Hi - I know this came up a little while ago while troubleshooting some access or downloads from a client.

While looking at our recent Grails upgrade, I generated a new grails app using the latest 5.2.1.  They still use `https://repo.grails.org/grails/core` in the generated build.gradle file, which makes me think we should probably stick with that as well.

Anything I am not/mis-remembering that would cause us to change our mind, or should we close this out for the moment?

PR #591, opened 2022-07-06
Ag grid28

lbwexler:


TomTirapani:
Looks good to me, seems to work fine from what I can see. Great that we're able to use modules to reduce the bundle size / time too.

TomTirapani (approved):

PR #595, opened 2022-08-03
Enable multi button group input

lbwexler:


client-app/src/mobile/buttons/ButtonPageModel.js
```diff
@@ -7,6 +7,7 @@ export class ButtonPageModel extends HoistModel {
     @bindable disabled = false;
     @bindable active = false;
     @bindable toolbar = false;
+    @bindable enablemulti = false;
```
lbwexler, line 10: should match case, e.g. `enableMulti`

PR #598, opened 2022-08-22
Add grid `sparklines` to Portfolio example

ghsolomon:
`LineChart.js` is no longer used.  Thought about deleting it, but I wonder if it still might be a valuable HighCharts example?

### DIFF
--- a/client-app/src/examples/portfolio/detail/OrdersPanelModel.js
+++ b/client-app/src/examples/portfolio/detail/OrdersPanelModel.js
@@ -1,7 +1,7 @@
 import {FilterChooserModel} from '@xh/hoist/cmp/filter';
 import {HoistModel, managed, XH} from '@xh/hoist/core';
 import {GridModel} from '@xh/hoist/cmp/grid';
-import {isNil} from 'lodash';
+import {isNil, map, uniq} from 'lodash';
 import {PERSIST_DETAIL} from '../AppModel';
 import {
     dirCol,
@@ -13,6 +13,8 @@ import {
     symbolCol, timeCol,
     traderCol
 } from '../../../core/columns';
+import {fmtDate} from '@xh/hoist/format';
+import {fmtNumber} from '@xh/hoist/format/FormatNumber';
 
 export class OrdersPanelModel extends HoistModel {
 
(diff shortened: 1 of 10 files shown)


client-app/src/examples/portfolio/detail/OrdersPanelModel.js
```diff
@@ -39,6 +41,28 @@ export class OrdersPanelModel extends HoistModel {
             persistWith: {...PERSIST_DETAIL, path: 'ordersGrid'},
             columns: [
                 {...symbolCol, pinned: true},
+                {
```
lbwexler, line 44: nit -- might put this in the column/positions file under the name `tradingVolSparklineCol` for clarity

client-app/src/examples/portfolio/detail/OrdersPanelModel.js
```diff
@@ -90,8 +114,12 @@ export class OrdersPanelModel extends HoistModel {
         }
 
         try {
-            const orders = await XH.portfolioService.getOrdersAsync({positionId, loadSpec});
-            gridModel.loadData(orders);
+            const orders = await XH.portfolioService.getOrdersAsync({positionId, loadSpec}),
+                symbols = uniq(map(orders, 'symbol')),
+                sparklineSeries = await XH.portfolioService.getSparklineSeriesAsync({symbols, loadSpec}),
+                rows = orders.map(order => ({...order, tradingVolume: sparklineSeries[order.symbol]}));
```
lbwexler, line 120: nit:  typically would just mutate the order, by installing the tradingVolume rather than spreading....

lbwexler:
So thought about this a bit more -- think you are right greg, that nice to keep the existing line chart example.  How about the following --

1) We restore the existing tab container with both charts as is.  This is a nice demo of tab container in this app, and multiple chart types.  Restore the back end single series endpoint

2) We add a dedicated endpoint that returns the *one month* price history in bulk, for the sparkline.  Can return, for example, just the close prices.  I think a seperate endpoint that is optimized for the bulk data loading of the limited time series is more in line with what we do in a real application and a nice pattern to show.  I also think the actual price might make a more appealing sparkline -- you could even do something like shading the area under the line green and red the way you see is so often done to indicate price movement above and below the initial starting point of the series.

ghsolomon:
> you could even do something like shading the area under the line green and red the way you see is so often done to indicate price movement above and below the initial starting point of the series.

I was hoping AG Grid would support this more easily out-of-the-box, but there doesn't appear to be a way to set the Y basis value.  Could normalize the data so the first price in the series is plotted at '0', and then make sure the tooltip shows the correct price.  For now, I just kept the sparkline as-is.  Can discuss more - might be a nice place where Hoist could provide some support to make things easier.

PR #633, opened 2022-12-23
Remove VersionBar from toolbox examples

Ryanseanlee:


lbwexler (commented):
I think this would be a nice small improvement if we actually returned false only if we are embedded in a frame (also a good example for apps that wanted to do something just like that.)

PR #644, opened 2023-03-12
Unpin logback version

lbwexler:


lbwexler:
not much here in this testing branch

PR #647, opened 2023-04-28
Added new declare statement in Bootstrap.ts, for IntelliJ

jskupsik:
This is the new "declare" statement that I came up with, also added comments to hopefully explain our magic.

I ended up keeping the `interface` for XHApi because it will otherwise give `error TS2323: Cannot redeclare exported variable`.

The `XH` const re-declaration is annotated with `// @ts-ignore` as to not give the `error TS2323` like above, but IntelliJ is able to notice uses of service methods better due to it.

Without the XH line:
image

With the XH line:
image

### DIFF
--- a/client-app/src/Bootstrap.ts
+++ b/client-app/src/Bootstrap.ts
@@ -14,9 +14,7 @@ import {OauthService} from './core/svc/OauthService';
 import {TaskService} from './examples/todo/TaskService';
 
 declare module '@xh/hoist/core' {
-    // @ts-ignore - A 'duplicate declaration' that helps IntelliJ better recognize uses of app services from the `XH` singleton.
-    export const XH: XHApi;
-    // Tells typescript to *statically* assume that these services/fields exist on the `XH` singleton.
+    // Merge interface with XHApi class to include injected services.
     export interface XHApi {
         contactService: ContactService;
         gitHubService: GitHubService;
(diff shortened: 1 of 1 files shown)


jskupsik:
Note to self, update the /hoist-react/ documentation with the new statement:
https://github.com/xh/hoist-react/blob/develop/docs/upgrade-to-typescript.md#bootstrapts--service-declarations

client-app/src/Bootstrap.ts
```diff
@@ -14,13 +14,17 @@ import {OauthService} from './core/svc/OauthService';
 import {TaskService} from './examples/todo/TaskService';
 
 declare module '@xh/hoist/core' {
+    // @ts-ignore - A 'duplicate declaration' that helps IntelliJ better recognize uses of app services from the `XH` singleton.
+    export const XH: XHApi;
+    // Tells typescript to *statically* assume that these services/fields exist on the `XH` singleton.
```
ghsolomon, line 19: I wonder if tweaking the comment a bit like this makes sense?

    // Merge interface with XHApi class to include injected services.

https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces

jskupsik: I like this better, thanks!

Issue #652, opened 2023-05-23
Improve/create fallback UI for Github and News panels when loaded without apiKeys in config


cnrudd:
The News demo app requires an api key to fetch news.
The Hoist Commits panel on the home page requires a GitHub Access Token to fetch commits from our repos.

As per security best practices, these keys/tokens are not coded into the [config defaults](https://github.com/xh/toolbox/blob/develop/grails-app/init/io/xh/toolbox/BootStrap.groovy#L79).  
So, if a developer runs the app with the H2 DB, these two sections of the app will silently load no data.

We should improve the UI in these two sections so that if no keys are in config (no matter the DB used), the UI shows a notification box that explains what config name with what kind of value is needed to get data.

PR #659, opened 2023-07-14
Add/improve explanatory text if unable to load news & github panels

connor15mcc:
As per #652, we can provide the user additional information when unable to load the news and github panes.

Suggestions welcome for the content of the text!

**This change depends on changes proposed in `hoist-react` [PR #3416](https://github.com/xh/hoist-react/pull/3416)**

image
image

### DIFF
--- a/client-app/src/desktop/tabs/home/widgets/activity/ActivityWidget.ts
+++ b/client-app/src/desktop/tabs/home/widgets/activity/ActivityWidget.ts
@@ -1,6 +1,6 @@
 import {grid} from '@xh/hoist/cmp/grid';
 import {gridCountLabel} from '@xh/hoist/cmp/grid/helpers/GridCountLabel';
-import {filler} from '@xh/hoist/cmp/layout';
+import {filler, p, vframe} from '@xh/hoist/cmp/layout';
 import {creates, hoistCmp} from '@xh/hoist/core';
 import {colChooserButton} from '@xh/hoist/desktop/cmp/button';
 import {filterChooser} from '@xh/hoist/desktop/cmp/filter';
@@ -9,14 +9,27 @@ import {panel} from '@xh/hoist/desktop/cmp/panel';
 import {toolbar, toolbarSep} from '@xh/hoist/desktop/cmp/toolbar';
 import './ActivityWidget.scss';
 import {ActivityWidgetModel} from './ActivityWidgetModel';
+import {errorMessage} from '@xh/hoist/desktop/cmp/error';
+import {XH} from '@xh/hoist/core/XH';
 
 export const activityWidget = hoistCmp.factory({
     displayName: 'ActivityWidget',
     model: creates(ActivityWidgetModel),
 
-    render() {
+    render({model}) {
+        let errorThrown = model.githubErrorThrown;
+        if (errorThrown && !errorThrown['isRoutine']) {
+            XH.handleException(errorThrown, {showAlert: false, showAsError: false});
+        }
+
         return panel({
-            item: grid(),
+            items: [
+                gitHubErrorPanel({error: errorThrown}),
+                vframe({
+                    omit: !!errorThrown,
+                    item: grid()
+                })
+            ],
             bbar: bbar(),
             mask: 'onLoad'
         });
@@ -45,4 +58,15 @@ const bbar = hoistCmp.factory({
             ]
         });
     }
+});
+
+const gitHubErrorPanel = hoistCmp.factory(({error}) => {
+    return errorMessage({
+        omit: !error,
+        error: error,
+        title: 'Error loading GitHub panel',
+        message: error?.['isRoutine']
+            ? p(error?.['message'])
+            : 'A non-routine exception occurred when loading this panel. Please try again later.'
+    });
 });
--- a/client-app/src/desktop/tabs/home/widgets/activity/ActivityWidget.ts
+++ b/client-app/src/desktop/tabs/home/widgets/activity/ActivityWidget.ts
@@ -1,6 +1,6 @@
 import {grid} from '@xh/hoist/cmp/grid';
 import {gridCountLabel} from '@xh/hoist/cmp/grid/helpers/GridCountLabel';
-import {filler, p, vframe} from '@xh/hoist/cmp/layout';
+import {filler} from '@xh/hoist/cmp/layout';
 import {creates, hoistCmp} from '@xh/hoist/core';
 import {colChooserButton} from '@xh/hoist/desktop/cmp/button';
 import {filterChooser} from '@xh/hoist/desktop/cmp/filter';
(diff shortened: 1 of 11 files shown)


client-app/src/desktop/tabs/home/widgets/activity/ActivityWidgetModel.ts
```diff
@@ -32,7 +32,19 @@ export class ActivityWidgetModel extends HoistModel {
         };
 
         this.gridModel = new GridModel({
-            emptyText: 'No commits found...',
+            // emptyText: 'No commits found...',
```
cnrudd, line 35: remove old code.

cnrudd (changes requested):
I think the new html is fine.  I might drop the "Maybe".

"Check that..." is crisper.

cnrudd (changes requested):
@connor15mcc No need to work on this over the weekend of course. I just wanted to put down these thoughts while I had them:

I think that the more appropriate change for these two panels would be to display the Hoist-React ErrorMessage component instead of the Dataview or Grid when we know that there is a missing config.

Toolbox has a small example that demos the ErrorMessage component here: https://toolbox.xh.io/app/other/errorMessage

So, the emptyText property could be left as is, and used only when data is really empty.

When the config is missing, an Error with the message you wrote could be created and set on the model.  This Error would then be picked up through context by the ErrorMessage component.

Toolbox's https://github.com/xh/toolbox/blob/master/client-app/src/desktop/tabs/other/ErrorMessagePanel.tsx
has a good example of switching off/on the display of "healthy" content/ErrorMessage.

client-app/src/desktop/tabs/home/widgets/activity/ActivityWidget.ts
```diff
@@ -46,3 +59,14 @@ const bbar = hoistCmp.factory({
         });
     }
 });
+
+const gitHubErrorPanel = hoistCmp.factory(({error}) => {
+    return errorMessage({
+        omit: !error,
+        error: error,
+        title: 'Error loading GitHub panel',
+        message: error?.['isRoutine']
```
connor15mcc, line 68: use the error type to dictate what to display to the user. If routine, we can feel free to expose the message content, otherwise provide a generic message

client-app/src/desktop/tabs/home/widgets/activity/ActivityWidgetModel.ts
```diff
@@ -167,13 +171,20 @@ export class ActivityWidgetModel extends HoistModel {
         });
 
         this.addReaction({
-            track: () => XH.gitHubService.allCommits,
+            track: () => [XH.gitHubService.allCommits, XH.gitHubService.errorThrown],
```
connor15mcc, line 174: Not quite sure if this is necessary, still getting used to mobx

client-app/src/desktop/tabs/home/widgets/activity/ActivityWidgetModel.ts
```diff
@@ -167,13 +171,20 @@ export class ActivityWidgetModel extends HoistModel {
...
 
     override async doLoadAsync() {
-        this.gridModel.loadData(XH.gitHubService.allCommits);
+        let allCommits = XH.gitHubService.allCommits,
+            thrownError = XH.gitHubService.errorThrown;
+
+        if (allCommits.length != 0) {
+            this.gridModel.loadData(allCommits);
+        } else if (thrownError) {
+            this.githubErrorThrown = thrownError;
```
connor15mcc, line 186: store any thrown errors on the model for reference in the component

connor15mcc:
After conversations with Lee and Colin, we've redefined the behavior of this feature. In general, the server is responsible for communicating missing config keys (by responding with non-200 status codes). The client then uses these codes to determine the severity of the corresponding exception, and possibly passing their messages onto the user.

The message from the 400 response is used to populate the ErrorMessage component (when a `routine` exception occurs):
image

grails-app/init/io/xh/toolbox/BootStrap.groovy
```diff
@@ -137,12 +137,6 @@ class BootStrap {
                     groupName: 'Toolbox - Example Apps',
                     note: 'Absolute path to disk location for storing uploaded files.'
             ],
-            newsApiKey: [
```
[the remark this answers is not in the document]
lbwexler: isn't it useful to have the key established, commented, and grouped here?

client-app/src/examples/news/AppComponent.ts
```diff
@@ -33,7 +39,18 @@ export const AppComponent = hoistCmp({
...
                 appMenuButtonProps: {hideLogoutItem: false}
             }),
-            item: newsPanel()
+            items: [newsErrorPanel({error: newsErrorThrown}), newsPanel({omit: !!newsErrorThrown})]
         });
     }
 });
+
+const newsErrorPanel = hoistCmp.factory(({error}) => {
+    return errorMessage({
```
lbwexler, line 48: nice pattern here in the local component is to early out on null, avoiding the null checks below.

client-app/src/examples/news/AppComponent.ts
```diff
@@ -1,18 +1,24 @@
...
+import {errorMessage} from '@xh/hoist/desktop/cmp/error';
 
 export const AppComponent = hoistCmp({
     displayName: 'App',
     model: uses(AppModel),
 
     render({model}) {
+        let newsErrorThrown = model.newsPanelModel.newsErrorThrown;
+        if (newsErrorThrown && !newsErrorThrown['isRoutine']) {
+            XH.handleException(newsErrorThrown, {showAlert: false, showAsError: false});
```
lbwexler, line 19: don't want to 'handle' exception in render method -- not looking to do any state changing methods during render cycle.  Move this handling to the model.

client-app/src/examples/news/NewsPanelModel.ts
```diff
@@ -47,8 +51,12 @@ export class NewsPanelModel extends HoistModel {
     }
 
     override async doLoadAsync(loadSpec) {
-        const stories = await XH.fetchJson({url: 'news', loadSpec});
-        this.completeLoad(stories);
+        try {
+            const stories = await XH.fetchJson({url: 'news', loadSpec});
+            this.completeLoad(stories);
+        } catch (e) {
+            this.newsErrorThrown = e;
```
lbwexler, line 58: this is where we want to handle the exception in the case of a routine error.

client-app/src/examples/news/NewsPanelModel.ts
```diff
@@ -9,8 +9,12 @@ import {newsPanelItem} from './NewsPanelItem';
 export class NewsPanelModel extends HoistModel {
     SEARCH_FIELDS = ['title', 'text'];
 
+    @bindable.ref
+    newsErrorThrown: object;
```
lbwexler, line 13: - Don't think `object` is the right type.  Also -- a nit -- `error` is probably sufficient for a name here.  `news` and `thrown` implied.

grails-app/services/io/xh/toolbox/app/GitHubService.groovy
```diff
@@ -157,6 +158,13 @@ class GitHubService extends BaseService {
         }
     }
 
+    Map<String, CommitHistory> getCommitsByRepo() {
+        if (configService.getString('gitHubAccessToken', 'none') == 'none') {
```
lbwexler, line 162: Clever use of API to find unset key,.

lbwexler:
Added some specific comments, but would like to go through this together.  I think the hoops we are jumping through to get definitive evidence of just the unset key down to the client is a mistake.   We should either cache any loading exception on the server and always throw it to the client, or continue to just return empty records to the client, and guess about the problem in our warning message.

Lots to learn on this one -- lets discuss in person -- apologies if it seems like I am running you in circles here.

connor15mcc:
Adjusted the text to be more vague about possible points of failure, as back to not storing them on the client or server. Instead, we provide a slight hint targeted towards those who are running toolbox locally without full configuration.

image

PR #666, opened 2023-08-01
Autosize broken for  Sample Tree Grid w/Checkbox

leogretz2:
Temporarily substituted a width of 240 to override the default Grid column width of 200. SampleTreeGridModel is used in a couple places where the grid does not have checkboxes but the added 40px does not make a difference. Once autosizing is fixed for the Tree grids with checkboxes, the width parameter can be removed from the name column altogether.


TomTirapani:
Thanks Leo, looks great and as discussed 👍 

Lee, for extra context, I explained that our preferred solution would be to leverage the autosize system with a min / max for this column - but unfortunately autosize is not accounting for the checkbox properly and will always truncate when autosized.

Having thought about this a bit more, I remembered we have this `autosizeBufferPx` property on Column:
https://github.com/xh/hoist-react/blob/develop/cmp/grid/columns/Column.ts#L313

This adds an additional width to whatever width the autosize is calculated to. We might be able to use this to workaround the autosize issue (which is a separate issue we should address, but not one with an immediate solution).

Here's what I recommend we do:
+ Add a suitable `autosizeBufferPx` to the column when `includeCheckboxes` is true. It may take some experimentation to get the right value, probably something like 40.
+ Add a suitable `autosizeMinWidth` and `autosizeMaxWidth` to the root `nameCol` to prevent is becoming unweildly.
+ Set the GridModel's `autosizeOptions.mode` to 'managed', so that it automatically autosizes the column. https://github.com/xh/hoist-react/blob/develop/cmp/grid/enums/GridAutosizeMode.ts#L32

I think with the above we can remove the default width from nameCol and leverage the autosizing system to display the column properly without truncating.

lbwexler:
@leogretz2 will fix up this PR, as it looks like it accrued a different change

I can create a new ticket with @TomTirapani comments on a deeper fix. Thanks guys

leogretz2:
I will fix the PR by removing the react-select related commits and then try @TomTirapani 's suggested changes.

lbwexler:
deleting the PR -- Leo will create a new one based on Tom's suggestions which are now in a hoist-react PR

PR #670, opened 2023-08-31
Multi instance

lbwexler:


grails-app/services/io/xh/toolbox/app/NewsService.groovy
```diff
@@ -20,15 +22,15 @@ class NewsService extends BaseService {
...
-        if (!newsTimer) {
-            newsTimer = createTimer(
-                    runFn: this.&loadAllNews,
-                    interval: 'newsRefreshMins',
-                    intervalUnits: MINUTES,
-                    runImmediatelyAndBlock: true
-            )
-        }
-        return _newsItems ?  _newsItems : Collections.emptyList()
+        newsTimer = newsTimer ?: createTimer(
```
ghsolomon, line 25: Flagging to discuss (related to our Timer conversation yesterday). I could be missing something.  It seems like if a non-primary node is hit first, `_newsItems` will be empty, the timer `runFn` will _not_ run, and this method will fail to return any news items.

lbwexler: correct -- this would benefit from waiting

ghsolomon: Yeah- except this timer is lazily created, so I believe it could end up waiting indefinitely (until the same method happens to be called on the primary node).

I have some ideas I'd love to discuss tomorrow if you've got time.

lbwexler:
sure -- looking forward to it!

PR #671, opened 2023-09-13
DRY with recently-added props Panel.collapsedTitle and Panel.collapsedIcon

saba-mo:
See hoist-react PR https://github.com/xh/hoist-react/pull/3478

### DIFF
--- a/client-app/src/examples/portfolio/GridPanelModel.ts
+++ b/client-app/src/examples/portfolio/GridPanelModel.ts
@@ -27,11 +27,8 @@ export class GridPanelModel extends HoistModel {
         return this.gridModel.selectedRecord;
     }
 
-    /** Render a title with groupBy dimensions, but only when collapsed (to aid user in re-expanding). */
-    get title() {
-        return this.panelModel.collapsed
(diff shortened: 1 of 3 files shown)


client-app/src/examples/portfolio/GridPanelModel.ts
```diff
@@ -6,7 +6,6 @@ import {PERSIST_MAIN} from './AppModel';
 import {mktValCol, nameCol, pnlCol} from '../../core/columns';
 import {PortfolioPanelModel} from './PortfolioPanelModel';
 import {capitalize} from 'lodash';
-import {Icon} from '@xh/hoist/icon';
 
 export class GridPanelModel extends HoistModel {
     @managed
```
TomTirapani, line 11: Now that we no longer need to externally observe its collapsed state, I think we could further clean this up by replacing this PanelModel with an inline `modelConfig: {}` in the component. This is a major benefit of supporting the collapsed title switching internally.

saba-mo: Good point! I've made the change.

PR #691, opened 2024-01-10
Assign local admin roles if needed

ghsolomon:


cnrudd (approved):
I confirm that this change fixes the error that prevents the server in the develop branch from starting.

But once started, the app has other errors, such as:
Screenshot 2024-01-10 at 5 04 37 PM

ghsolomon:
CR Lee

PR #692, opened 2024-01-10
Move createLocalAdminUserIfNeeded to before services init

cnrudd:
& Move jsLicenses config into ensureRequiredConfigsCreated

I run with H2, so the DB gets recreated every time.  Maybe that helped find this issue?

Without this fix, user.hasRole(..  in RoleAdminService.ensureUserHasRoles throws an NPE because there is no user.

and then this error shows up in the UI:
Screenshot 2024-01-10 at 5 04 37 PM


ghsolomon:
Good catch!

amcclain:
@cnrudd meant to also say thanks for noting and fixing 👍

PR #696, opened 2024-02-05
Role refactor

lbwexler:


amcclain:
@lbwexler let's touch base on this tomorrow - we can go over `DefaultRoleEditService` vs. `DefaultRoleAdminService` in hoist-core and take a final decision on which survives 😄

lbwexler:
Hah!   Default RoleUpdateService might have gone over better and could be
the answer.  Like having it in a separate service for implementation
purposes -- we could delegate to it and make DefaultRoleAdminService the
main interface.  Wish we had a more well-defined pattern for implementation
services.

PR #700, opened 2024-03-07
Removes adminUsername usage in favor of bootstrap, adds default Auth0 configs

minusleaf:
- Added a comment where the key is added in case people look through the code and have their mental alarm bells ring.


lbwexler:
Don  -- went ahead and made a few additional changes on the client, to adopt the common Auth0 always approach, and DRY things up, and also updated the readme.  One question we still have is, whether we need a non-Auth0 approach (for when we have config issues, or are in `runOnDevice` mode.

client-app/src/BaseAppModel.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    override async logoutAsync() {
+        await XH.oauthService.logoutAsync();
+    }
+
+    override get supportsVersionBar(): boolean {
+        return window.self === window.top;
+    }
+
+    static async isSessionActiveOnDev(): Promise<boolean> {
```
amcclain, line 20: This name doesn't feel quite right to me - would something like `shouldSkipOauth` be a bit more expressive? I'd also add a comment to explain why we have this check, as it's a subtle issue and I don't think I would remember 😄 .

I see that the bit about the delay was copied here from the AppContainerModel implementation. That would not apply to Toolbox - believe it's for alternate SSO-ish flows where the user might be presented with a browser credentials prompt due to a filter-based auth negotiation.

client-app/src/core/svc/OauthService.ts
```diff
@@ -42,15 +42,11 @@ export class OauthService extends HoistService {
...
-            XH.appSpec.isSSO = false;
-            return;
-        }
-
         if (!config?.domain || !config?.clientId) {
             throw XH.exception(`
                 Unable to init OAuthService - expected config not returned by server. 
-                Please log on as a local admin user and review this instance's soft configuration entries.
+                Please review the settings in configs "auth0ClientId", "auth0Domain",
+                and "auth0Jks".  Default values for these configs are provided in "Bootstrap.groovy"
```
amcclain, line 49: We could remove the mention of auth0Jks - I don't think that's quite the right config name, and it's not what's being checked here. I'm on the fence about calling out Bootstrap.groovy here in this error message, but since it would almost certainly be encountered by a developer I do think it might be useful to be that specific.

client-app/src/desktop/AppModel.ts
```diff
@@ -44,18 +44,8 @@ export class AppModel extends HoistAppModel {
...
-    static override async preAuthAsync() {
-        const whitelist = ['toolbox.xh.io', 'toolbox-dev.xh.io', 'localhost'];
-        if (whitelist.includes(window.location.hostname)) {
-            await XH.installServicesAsync(OauthService);
-        } else {
-            // We are not running in an environment that is supported by auth0
-            // (presumably `yarn startOnDevice`), so fall back to username / password prompt.
-            XH.appSpec.isSSO = false;
-        }
-    }
```
amcclain, line 56: I apologize, but this bit where we detect the common case where we don't expect Auth0 to work (non-whitelisted hostname) and automatically fallback to the form-based login is quite nice - I had forgotten we had this in here and that it provided a pretty good solution to the "startOnDevice" problem (although not the "auth0 is misconfigured" problem).

Could we mix this in to the new preAuth check, so we get the best of all worlds?  That way a user who is running on a non-standard hostname and has the bootstrapAdminUser set could get the working form-based auth directly from the app they meant to load.  The bootstrapLogin would then be a true backup in the event that auth0 is broken.

client-app/src/examples/filemanager/AppModel.ts
```diff
@@ -1,18 +1,5 @@
...
-        await XH.installServicesAsync(OauthService);
-    }
-
-    override async logoutAsync() {
-        await XH.oauthService.logoutAsync();
-    }
-
-    override get supportsVersionBar(): boolean {
-        return window.self === window.top;
-    }
```
amcclain, line 17: Great that you added the BaseAppModel - nice to see this duplicated code go away.

grails-app/init/io/xh/toolbox/BootStrap.groovy
```diff
@@ -77,6 +76,43 @@ class BootStrap {
...
+                                            ]
+                                    ]
+                            ]
+                    ],
+                    clientVisible: false,
+                    groupName: 'Auth0',
+                    note: 'JSON Web Key Set (JWKS) for validating Auth0 ID tokens.'
```
amcclain, line 114: These are sourced from https://login.xh.io/.well-known/jwks.json - we could add that URL to this comment to help document.  Even better we could "do the right thing" and load+cache the JWKS in our service so we don't need it in the config at all.  At this point that seems like it would be a worthy exercise.

(PS see https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets for some general info from Auth0 on JWKS)

grails-app/services/io/xh/toolbox/user/RoleService.groovy
```diff
@@ -15,21 +15,10 @@ class RoleService extends DefaultRoleService {
 
     void init() {
         super.init()
-        assignLocalAdminRolesIfNeeded()
     }
```
amcclain, line 18: Let's remove the init() method here entirely, since we have no unique work to do in it other than call super

amcclain (commented):
Thanks for working through all of this!  It's a combination of funny cases, for sure, but this will be a nice clarification.  Some comments below for your review, happy to discuss whenever we have a chance.

client-app/src/BaseAppModel.ts
```diff
@@ -0,0 +1,38 @@
+import {HoistAppModel, XH} from '@xh/hoist/core';
+import {OauthService} from './core/svc/OauthService';
+
+export class BaseAppModel extends HoistAppModel {
+    static override async preAuthAsync() {
+        if (!(await BaseAppModel.shouldSkipOAuth())) {
```
lbwexler, line 6: might be slightly clearer to have OAuthService do this logical check internally in its init() and become `disabled` and be a no-op.  Situates that common code in a natural place and also avoids the conditional presence of the service.

lbwexler: could also just have the app immediately go in to "non-sso" mode the way we did before, if oauth is disabled 

XH.appSpec.isSSO = false;

That way, we don't need any special app in the skip case!

client-app/src/core/svc/OauthService.ts
```diff
@@ -101,7 +97,7 @@ export class OauthService extends HoistService {
         return claims?.__raw;
     }
 
-    private async checkAuthAsync() {
+    private async checkAuthAsync(): Promise<boolean> {
```
lbwexler, line 100: nice!

grails-app/services/io/xh/toolbox/security/Auth0Service.groovy
```diff
@@ -17,17 +20,13 @@ import static io.xh.hoist.util.InstanceConfigUtils.getInstanceConfig
...
 class Auth0Service extends BaseService {
 
+    private JSONClient _jsonClient
 
-    static clearCachesConfigs = ['auth0Domain', 'auth0ClientId', 'auth0Jwks']
+    static clearCachesConfigs = ['auth0Domain', 'auth0ClientId']
 
     ConfigService configService
 
     Map getClientConfig() {
```
lbwexler, line 29: it occurs to me that, we could actually even just go back to having the server disable this if the bootstrap user is set, and then not worry about any of the other logic!  Or even better do both -- e.g. throw here if no bootstrapUser set and  domains are not in the whitelist that will allow oAuth to work....

PR #702, opened 2024-03-13
Uses Hoist-Dev-Utils and Hoist-React branches that have new chunking strategy.

cnrudd:


lbwexler:
@cnrudd -- is this obsolete?

Issue #458, opened 2020-12-21
Routing examples


amcclain:
Would like to have examples of routing baked into Toolbox so we can look to it for a reference implementation.

I would suggest having two new tabs under "Other" - 

### Routing (simple)

* Grid with the selected record ID maintained in the route. E.g. URLs like `https://toolbox-dev.xh.io/app/other/simpleRouting/123` where `123` is a record ID and that record is (auto) selected in the grid.

### Routing (advanced)

* Grid with query params (e.g. use of groupingChooser and/or filters) where query and selected ID are encoded into the route. In this case we could demo the pattern of writing route state into a JSON object then Base64 encoding to make the route param itself an opaque blob. E.g. URLs like `https://toolbox-dev.xh.io/app/other/advancedRouting/eyJncm91cEJ5IjoidHJhZGVyIiwic2VsZWN0ZWRJZCI6MTIzfQ==`, where the routing param decodes to `{"groupBy":"trader","selectedId":123}`.

PR #705, opened 2024-03-15
Advanced Routing Panel in "Other" tab

jacob-xhio:
This PR addresses the advanced case of the following issue: https://github.com/xh/toolbox/issues/458. A video sample of the functionality working is shown below, along with a screenshot of the final appearance of the tab.

https://github.com/xh/toolbox/assets/163027616/6b1202c8-ede9-4360-9162-6e1a68ccb124

Screenshot 2024-03-15 at 12 10 32 PM


client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,171 @@
...
+        if (selectedId) await this.setSelected(selectedId);
+    }
+
+    @action
+    private updateRoute() {
+        if (!XH.routerState.name.startsWith('default.other.advancedRouting')) return;
+        const {groupBy, sortBy} = this;
+        const selectedId = this.gridModel.selectedRecord?.id;
+        const routeParam = btoa(JSON.stringify({groupBy, sortBy, selectedId}));
+        XH.navigate('default.other.advancedRouting.routeParam', {routeParam});
```
haynesjm42, line 163: Probably want `{replace: true}` as the third arg so avoid extra history entries

haynesjm42 (commented):
Some additional things to consider for our advanced routing example(s):

--------------------

A pattern I've been using for the obfuscated/encoded route params is this:

Using these encoder/decoder methods on the route to automatically handle the encoding without needing to do anything in the tab/component.

```
// Encoding of json route params as base64
export const routeParamEncoders = {
    encodeParams: params => {
        if (isEmpty(params)) return {};
        return {q: window.btoa(JSON.stringify(params))};
    },
    decodeParams: params => {
        if (!params.q) return {};
        return JSON.parse(window.atob(params.q));
    }
};
```

This also requires putting the router query params in "loose" mode (which just means you can put whatever you want in the params and router5 wont validate/modify/etc.)

`XH.router.setOption('queryParamsMode', 'loose');`

Then in the actual route definition you can do this:

```
...

{
    name: 'myRoute',
    path: '/my-route',
    ...routeParamEncoders
},
...
```

With this your component code doesn't need to worry about the encoding at all, or even what the route is configured to use for params - it just gets the json object params and puts whatever json object it wants into the params.

--------------------

Another thing that can be important for tabs which allow you to modify something in a non-modal way is to add handlers to intercept deactivation of a route so you can prompt the user to save changes, which can be done like this:

```
XH.router.canDeactivate(route, () => (toState, fromState, doneFn) => {
    if (this.isModified) {
        XH.confirm(...).then(confirmed => {
            if (confirmed) doneFn();
        }
    } else {
        doneFn();
    }
});
```

Also note that this apparently needs to be added each time the route is activated, so it should go into a reaction on the route after checking that the current route is the one for your tab/component (just like when parsing route params - would go in the same place)

This is usually done in combination with adding a `beforeunload` handler on the window to make a similar check for avoiding losing unsaved changes.

```
window.addEventListener('beforeunload', e => {
    // Ignore if we aren't the current page or if we don't have pending changes
    if (!XH.routerState.name.endsWith('myRoute') || !this.isModified) {
        delete e.returnValue;
        return;
    }

    e.preventDefault();
    e.returnValue = '';
});
```

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
jacob-xhio: Implemented this selectively -- will replace when a new record is selected in the grid, but will navigate when grouping or sorting is changed.

client-app/src/desktop/AppModel.ts
```diff
@@ -21,7 +21,19 @@ import {panelsTab} from './tabs/panels/PanelsTab';
 import {fmtDateTimeSec} from '@xh/hoist/format';
 import {span} from '@xh/hoist/cmp/layout';
 import {BaseAppModel} from '../BaseAppModel';
+import {isEmpty} from 'lodash';
 
+// Encoding of json route params as base64
+export const routeParamEncoders = {
```
amcclain, line 27: This is kind of a funny thing to now have at the very top of our primary desktop AppModel, in terms of the being the first thing a reader sees.  Could they be moved to a utils file, or even just down near the bottom of this file?

client-app/src/desktop/AppModel.ts
```diff
@@ -47,6 +59,7 @@ export class AppModel extends BaseAppModel {
     override async initAsync() {
         await super.initAsync();
         await XH.installServicesAsync(GitHubService, PortfolioService);
+        XH.router.setOption('queryParamsMode', 'loose');
```
amcclain, line 62: What's this do? Could you add a comment? Easy to miss if it's important to the functioning of the example (which I assume it is)

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+                    decoded and parsed to restore the state.
+                </p>,
+                <p>
+                    The current state encoding is: <br />
+                    <br />
+                    <code>groupBy: {model.groupBy || 'None'}</code>
+                    <br />
+                    <code>sortBy: {model.sortBy || 'None'}</code>
+                    <br />
+                    <code>selectedId: {model.gridModel.selectedRecord?.id || 'None'}</code>
```
amcclain, line 44: I had expected this little output to be showing the decoded route params more directly, rather than relaying the state from the model itself.  I think that would be a bit more interesting / direct, to show the decoded route as it exists for the current page

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+                    <br />
+                </p>,
+                <p></p>
+            ],
+            item: panel({
+                ref: model.panelRef,
+                mask: 'onLoad',
+                item: hframe(
```
amcclain, line 52: No need for this hframe - it's designed to layout two child components side-by-side horizontally, but we don't have that here.

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+                    vframe(
+                        grid(),
+                        hbox({
+                            items: [Icon.info()],
+                            className: 'tb-sample-grid__selbar'
```
amcclain, line 57: What's this doing?

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+                        })
+                    )
+                ),
+                tbar: [
+                    refreshButton(),
+                    colAutosizeButton(),
```
amcclain, line 63: We could lose this - doesn't seem relevant

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+                            {value: 'city|desc', label: 'City (Desc)'},
+                            {value: 'city|asc', label: 'City (Asc)'},
+                            {value: 'trade_date|desc', label: 'Trade Date (Desc)'},
+                            {value: 'trade_date|asc', label: 'Trade Date (Asc)'},
+                            {value: null, label: 'None'}
+                        ]
+                    }),
+                    switchInput({
+                        label: 'Prevent Route Deactivation',
+                        onChange: () => (model.preventDeactivate = !model.preventDeactivate)
```
amcclain, line 92: Use `bind: 'preventDeactivate'` instead

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+                ]
+            })
+        });
+    }
+});
+
+class AdvancedRoutingPanelModel extends HoistModel {
+    @observable groupBy = null;
+    @observable sortBy = null;
+    @observable preventDeactivate = false;
```
amcclain, line 105: make this `@bindable`

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+    constructor() {
+        super();
+        makeObservable(this);
+
+        this.addReaction({
+            track: () => XH.routerState.params,
+            run: () => this.parseRouteParams(),
```
amcclain, line 113: Would it be helpful to adopt the same convention between this model and the simple example, in terms of the route->state and state->route method names?

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+        window.addEventListener('beforeunload', e => {
+            if (!XH.routerState.name.startsWith('default.other.advancedRouting')) {
+                delete e.returnValue;
+                return;
+            }
+            if (this.preventDeactivate) e.preventDefault();
+        });
+    }
+
+    gridModel = new GridModel({
```
amcclain, line 132: We'd typically either set this within the ctor or (if very simple as it is) above

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+        columns: [
+            {field: 'id', flex: 0},
```
amcclain, line 134: no flex:0

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+            {field: 'city', flex: 1},
+            {field: 'trade_date', flex: 1}
+        ]
+    });
+
+    @action
+    private setGroupBy(groupBy: string) {
+        this.groupBy = groupBy;
+
+        // Always select first when regrouping.
```
amcclain, line 145: ?

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+    private async setSelected(recordId: string | number) {
+        await this.gridModel.selectAsync(Number(recordId));
+        if (!this.gridModel.selectedId) {
+            XH.dangerToast(`Record ${recordId} not found`);
+        }
+    }
+
+    @action
+    private async parseRouteParams() {
+        if (this.gridModel.empty) return;
```
amcclain, line 169: Necessary?

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+            this.gridModel.empty
+        )
+            return;
+        const {groupBy, sortBy} = this;
+        const selectedId = this.gridModel.selectedRecord?.id;
+        XH.navigate(
+            'default.other.advancedRouting',
+            {groupBy, sortBy, selectedId},
+            // Only push URL to route history if groupBy or sortBy changes.
+            {replace: selectedId != XH.routerState.params.selectedId}
```
amcclain, line 189: Nice!

amcclain (commented):
Again, a few small comments but looking very close / let's get it in today 👍

client-app/src/desktop/AppModel.ts
jacob-xhio: Couldn't find a relevant existing Utils file to add to, so just moved to the bottom.

client-app/src/desktop/AppModel.ts
jacob-xhio: Per John's comment above: 

``` 
This also requires putting the router query params in "loose" mode (which just means you can put whatever you want in the params and router5 wont validate/modify/etc.)

XH.router.setOption('queryParamsMode', 'loose');
```

Will add comment above this line to explain in the code.

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
jacob-xhio: Adds that bar in the bottom with the "info" icon for aesthetic purposes. Will remove as it is not pertinent to demonstrating advanced routing.

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
jacob-xhio: This is necessary because `parseRouteParams` is ran when the page initializes, before the grid model loads in its data. When removed, this will cause the state passed in via `routeParam` to be reset to the no-selection case.

client-app/src/desktop/tabs/other/routing/AdvancedRoutingPanel.tsx
jacob-xhio: Removed comment, was a vestige of copy-pasta.

amcclain:
If I open the advanced example, adjust some params, then swap to another tab, I see the decoded route params now in my URL:
Screenshot 2024-03-26 at 4 32 52 PM

amcclain:
@jacob-xhio - it looks like the "revent route deactivation" switch is not working, or at least I could not work out what it was supposed to do while testing out the app.  Have a look please?

PR #710, opened 2024-04-10
Update tsconfig.json to ref hoist-react as composite build

amcclain:
+ Allow for distinct tsconfig options in HR vs. app-level. Relies on `composite: true` flag to be set in hoist-react tsconfig.json.
+ Re-order tsconfig options for clarity.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ bin/
 out/
 /client-app/build/
 /client-app/node_modules/
+/client-app/tsconfig.tsbuildinfo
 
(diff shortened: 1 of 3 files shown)


amcclain:
For merging after https://github.com/xh/hoist-react/pull/3627

amcclain:
See additional explanatory comment on https://github.com/xh/toolbox/pull/710/commits/3e0763ee76730ae181ac23da8e4c0339b1d49bf6

PR #712, opened 2024-04-23
Added tags data/column to TradeService and ColumnFilteringPanel grid

jskupsik:


jskupsik:
hoist-react PR: https://github.com/xh/hoist-react/pull/3637

PR #721, opened 2024-05-22
Assets, configs, and docs for running with HTTPS

cnrudd:


README.md
```diff
@@ -88,6 +88,26 @@ When adding a new top-level entry-point for Toolbox (such as a new example appli
...
+-Dserver.ssl.enabled=true
+-Dserver.ssl.certificate=classpath:local-dev/toolbox-local.xh.io.crt
+-Dserver.ssl.certificate-private-key=classpath:local-dev/toolbox-local.xh.io.key
+-Dserver.ssl.trust-certificate=classpath:local-dev/toolbox-local.xh.io.crt
+```
+3. Visit `https://toolbox-local.xh.io:8443/ping` in your browser to proceed past the SSL warning 
+  for API calls.
+4. Start the GUI with the `startWithHoistSecure` npm script.  Go to `https://toolbox-local.xh.io:3443/app/home` 
+   in your browser and proceed past the SSL warning.
+
```
amcclain, line 109: Thanks for baking these instructions into the README, really appreciate it.

Would we ever foresee needing to regen the bundled certs - e.g. do they ever expire?  If so and if you have the commands to generate handy, might want to copy those in here as well.  But truly no need if they should be stable or the recipe is sufficiently obvious.

src/main/resources/local-dev/toolbox-local.xh.io.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
amcclain, line 1: We can see if this gets flagged by any automated secret sniffing routines.  

Wondering if we should put a word like "test" or "self-signed" in the filenames to help drive home the point that these are not real credentials.  I know it says "local" maybe that's enough.  (To be clear, don't think we need the test domain name to be any different, just talking about these three filenames)

amcclain (approved):
Thanks for putting this together - all for merging pls see/resolve README comment then 👍 for merging from me.

README.md
```diff
@@ -88,6 +88,26 @@ When adding a new top-level entry-point for Toolbox (such as a new example appli
...
+especially when testing OAuth, CORS, or cookie dependent features. 
+Follow these steps to run with HTTPS on the `toolbox-local.xh.io:3443` domain:
+1. add this entry to your dev machine's `hosts` file: `127.0.0.1 toolbox-local.xh.io`
+2. start the Grails server with the following VM Options (the referenced files are self-signed certs commited to the repo for local dev purposes):
+```
+-Dserver.port=8443
+-Dserver.ssl.enabled=true
+-Dserver.ssl.certificate=classpath:local-dev/toolbox-local.xh.io.crt
+-Dserver.ssl.certificate-private-key=classpath:local-dev/toolbox-local.xh.io.key
+-Dserver.ssl.trust-certificate=classpath:local-dev/toolbox-local.xh.io.crt
```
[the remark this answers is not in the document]
cnrudd: fixed

README.md
cnrudd: extended exp. to 10 years from now.   Not sure exp. matters, since already invalid.  I have a small npm node script used to make the certs.  happy to show you, discuss whether worth putting in its own repo.

src/main/resources/local-dev/toolbox-local.xh.io.key
cnrudd: added 'self-signed' to file names.

cnrudd:
a couple thoughts:
the Grails server port change to 8443 is not strictly needed.  SSL could run on 8080.
The webpack config: --env baseUrl=https://toolbox-local.xh.io:8443/  could be --env baseUrl=https://toolbox-local.xh.io:8080/

and then would only be necessary because we need to get the baseUrl value to use https, not http.

We could make HDU smarter to use protocol https if devHttps=true, and then that baseUrl arg could be left out.

The arg --env devWebpackPort=3443 is not necessary either, could be left at default 3000, but easier to distinguish with 3443, maybe.

cnrudd:
> a couple thoughts: the Grails server port change to 8443 is not strictly needed. SSL could run on 8080. The webpack config: --env baseUrl=https://toolbox-local.xh.io:8443/ could be --env baseUrl=https://toolbox-local.xh.io:8080/
> 
> and then would only be necessary because we need to get the baseUrl value to use https, not http.
> 
> We could make HDU smarter to use protocol https if devHttps=true, and then that baseUrl arg could be left out.
> 
> The arg --env devWebpackPort=3443 is not necessary either, could be left at default 3000, but easier to distinguish with 3443, maybe.

With this [commit in HDU](https://github.com/xh/hoist-dev-utils/commit/e647df1431f271842f7e27099437d2fdf864b8ae), I was able to simplify the SSL config a little.

PR #729, opened 2024-08-14
Add example for `markdown` component

amcclain:
+ Test out ability to import markdown files from the source code - ensure that works through the webpack build.


amcclain:
A bit ugly, and missing the "links" feature that Toolbox examples support (and which would be good to configure), but wanted to get this in before it was pre-empted by other tasks.

Motivated by desire to see if we could import a markdown file - we have a use case in a client app - and then while testing that out figured we might as well end up with a TB example of this newer component.

amcclain:
Screenshot 2024-08-14 at 3 37 32 PM

PR #733, opened 2024-09-04
Persistence Manager

Ryanseanlee:


client-app/src/examples/portfolio/PortfolioPanel.ts
```diff
@@ -1,18 +1,29 @@
...
 import {mapPanel} from './MapPanel';
 import {PortfolioPanelModel} from './PortfolioPanelModel';
+import {persistenceManager} from '@xh/hoist/desktop/cmp/persistenceManager';
 
 export const portfolioPanel = hoistCmp.factory({
     model: creates(PortfolioPanelModel),
 
     render() {
         return panel({
+            tbar: tbar(),
```
ghsolomon, line 16: Since the tbar component is so minimal and doesn't react to any state changes, I would just specify the tbar items inline (i.e. `tbar: [persistenceManager()]`)

client-app/src/examples/portfolio/PortfolioPanel.ts
```diff
@@ -1,18 +1,29 @@
...
             mask: 'onLoad',
```
ghsolomon, line 17: Probably want to mask while the `PersistenceManagerModel` is loading views for the first time. Otherwise, this will render unmasked with whatever the default view is (in code) and then jump to the persisted view once it loads

client-app/src/examples/portfolio/PortfolioPanelModel.ts
```diff
@@ -21,16 +30,40 @@ export class PortfolioPanelModel extends HoistModel {
     constructor() {
         super();
         const wsService = XH.webSocketService;
+
+        this.persistenceManagerModel = new PersistenceManagerModel({
+            type: 'portfoioExample',
+            noun: 'portfoio',
```
ghsolomon, line 36: sp: `portfolio` :)

client-app/src/examples/portfolio/PortfolioPanelModel.ts
```diff
@@ -28,6 +34,21 @@ export class PortfolioPanelModel extends HoistModel {
...
+        return this._initAsync()
+            .catch(e => {
+                XH.handleException(e, {showAlert: false});
+                this.initError = e;
+                throw e;
+            })
+            .linkTo(this.initTask);
+    }
+
+    private async _initAsync() {
```
ghsolomon, line 51: I think our convention here would be to call this `doInitAsync` rather than the underscore (that tends to be used more for private properties than methods).  But also curious why split this out at all.  Seems like it could all just go in `initAsync`

client-app/src/examples/portfolio/PortfolioPanelModel.ts
ghsolomon, line 1: In general, this is an annoying amount of boilerplate you had to write! I wonder if there'd be a way to generalize this and build a higher order component / model to avoid some of this.  Just brainstorming - what if the `PersistenceManager` component was a container that accepted a single child component that it would not render until it was done initializing.  Big issue here is that ideally the child would `create` its model, so that it would not be constructed until the manager finished initializing. BUT we would need a reference to that model in order to wire it.  Wonder if we could use `modelRef` to do that... but we wouldn't want whatever we do to be even more complicated / harder to reason about than what you have here...

Sorry for the flight of ideas - will think more on this, and maybe we can brainstorm together.

amcclain:
Picking this up with the latest WIP Hoist React APIs @ #737

PR #750, opened 2025-01-03
Added composite build settings.gradle file for running hoist inline

jskupsik:
Added a `settings.gradle` file to the toolbox app.

It is capable of running toolbox as a normal hoist app, in addition to being able to run toolbox with local hoist using the gradle composite build (`includeBuild`).

With this, the wrapper directory containing both toolbox and hoist-core no longer needs to have gradle installed nor have its own `settings.gradle` file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -43,19 +43,10 @@ springBoot {
     mainClass = xhAppPackage + ".Application"
 }
 
-if (parseBoolean(runHoistInline)) {
-    println "${xhAppName}: running with Hoist Core INLINE...."
-    grails.plugins {
-        implementation project(":hoist-core")
-    }
-} else {
-    println "${xhAppName}: running with Hoist Core PACKAGED at v${hoistCoreVersion}...."
-    dependencies {
-        implementation "io.xh:hoist-core:$hoistCoreVersion"
-    }
-}
-
 dependencies {
+    // If `runHoistInline` is true, `hoistCoreVersion` will be ignored and local hoist-core will be used instead.
+    implementation "io.xh:hoist-core:$hoistCoreVersion"
+
     // For server-side JWT validation.
     implementation "org.bitbucket.b_c:jose4j:0.9.6"
 
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,0 +1,11 @@
+import static java.lang.Boolean.parseBoolean
+
+rootProject.name = 'toolbox'
+
+// Composite build setup for compiling hoist-core locally
+if (parseBoolean(runHoistInline)) {
+    println "${xhAppName}: running with Hoist Core INLINE...."
+    includeBuild '../hoist-core'
+}else {
+    println "${xhAppName}: running with Hoist Core PACKAGED at v${hoistCoreVersion}...."
+}


amcclain:
I am very eager to get this merged - in fact I would like to use it more broadly across our apps at clients as well, where I am feeling the pain of not having wrapper projects setup to support development of client-specific plugins.

* Need to update with latest
* Need to update build in Teamcity:
Screenshot 2025-03-18 at 11 33 43 AM
* Need to communicate with team.

@lbwexler are you good with this change?
@jskupsik if Lee is on board, could you please do the above work so we can use this great change?

client-app/package.json
```diff
@@ -12,9 +12,11 @@
     "lint:js": "eslint --ext .js,.jsx,.ts,.tsx .",
     "lint:styles": "stylelint \"**/*.s?(a|c)ss\"",
     "start": "yarn install && cross-env NODE_OPTIONS=--max_old_space_size=3072 webpack-dev-server",
+    "start2": "yarn start --env devGrailsPort=8082 devWebpackPort=3002",
```
RyanMG, line 15: What are these updates doing? Does this version needs it own port- meaning will 8080 also be in use when running in this manner?

lbwexler: yes, indeed, this is for running a second client in development that will hit the second server

RyanMG: Thanks for the update. Perhaps a name update to something like `startSecondClient` / `startSecondClientWithHoist`?

lbwexler: Agree the name 2 sounds like a potential replacement!  Great catch.  But I suggest just changing out the '2' with `instance2` e.g. `startInstance2`, `startInstance2WithHoist

README.md
```diff
@@ -7,14 +7,13 @@ usage, and links to related Hoist and Toolbox source code. Toolbox also provides
...
-[Hoist React](https://github.com/xh/hoist-react) repos for detailed information and documentation on
-Hoist, including more information on configuring a local development environment.
+Please refer to the [Hoist Core](https://github.com/xh/hoist-core) and [Hoist React](https://github.com/xh/hoist-react) repos for detailed information and
+documentation on Hoist, including more information on configuring a local development environment.
 
 Toolbox is itself a Hoist Application, and we keep it updated with the latest versions of the Hoist
-Framework. A [toolbox-dev instance](https://toolbox-dev.xh.io) is auto-deployed via Teamcity on each
-commit to either the Toolbox or Hoist `develop` branches. We update a distinct
-["production" instance](https://toolbox.xh.io) manually with each new versioned Hoist release.
+Framework. A [toolbox-dev instance](https://toolbox-dev.xh.io) is auto-deployed via Teamcity on each commit to either the
```
RyanMG, line 14: Not your change, but link `Teamcity` to a [remote source](https://www.jetbrains.com/teamcity/?source=google&medium=cpc&campaign=AMER_en_US-PST+MST_TeamCity_Branded&term=teamcity&content=513031283514&gad_source=1&gclid=Cj0KCQjws-S-BhD2ARIsALssG0YIFLHx7-SjM5spUH4tV0PH1TguOVkh9zVY0o7RhZmdDntgPqwiEv4aAuf6EALw_wcB? )? Maybe most people know what this is, but I was only previously aware of Jenkins, Bamboo, Travis, CircleCI and AWS CodePipeLine. When I first read this I was unsure what Teamcity meant here.

README.md
```diff
@@ -64,59 +63,54 @@ When adding a new top-level entry-point for Toolbox (such as a new example appli
...
+
+* Create or find a suitable directory for multiple project repositories, like homedir or (if using
+  IntelliJ) `~/IdeaProjects/`.
+* Within this directory, check out the `toolbox`, `hoist-react`, and `hoist-core` repositories as
+  siblings.
+
+### Editing multiple projects together
+
+For editing Hoist Core and React alongside Toolbox, it is recommended to open the `hoist-react` and
+`hoist-core` projects as modules in your editor. Having all three repos in a single IntelliJ project
```
RyanMG, line 80: Assume IJ for the IDE. Do we want to present this README with this assumption?

build.gradle
```diff
@@ -43,19 +43,10 @@ springBoot {
     mainClass = xhAppPackage + ".Application"
 }
 
-if (parseBoolean(runHoistInline)) {
```
RyanMG, line 46: Personal ignorance only - what does `settings.gradle` do which we did not get here?

settings.gradle
```diff
@@ -0,0 +1,11 @@
+import static java.lang.Boolean.parseBoolean
+
+rootProject.name = 'toolbox'
+
+// Composite build setup for compiling hoist-core locally
+if (parseBoolean(runHoistInline)) {
+    println "${xhAppName}: running with Hoist Core INLINE...."
+    includeBuild '../hoist-core'
```
RyanMG, line 8: If we are including static pathing, we should 100% ensure the documentation denotes where things should live. **_It may_**, I just want to confirm we are thinking about this.

README.md
```diff
@@ -64,59 +63,54 @@ When adding a new top-level entry-point for Toolbox (such as a new example appli
...
+
+* To run the server using the local `hoist-core`, you need `hoist-core` to exist as sibling of the
+  `toolbox` package, and do one of the following:
+  * Edit the `toolbox/gradle.properties` file and set `runHoistInline=true`.
+  * Run the command `./gradlew bootRun -PrunHoistInline=true` from the `toolbox` directory.
+  * If using IntelliJ, run the `BootrunWithHoist` saved run configuration.
+
+  Note this is _only_ required if you're changing hoist-core code.
+
+* To run the client using the local `hoist-react`, you need `hoist-react` to exist as sibling of the
```
RyanMG, line 103: An extension to what I have already said, but there is a lot of details here (and previously) about second tier setups (including hoist in local dev), but it maybe misses the quick, happy path setup. 

I think this is ok for now, but it may be worth a think about how we present setup to a new user, and tier the information they receive - from quick start happy path to more in depth configurations as they need it.

jskupsik: Added some (at least bare bones) info about the happy path above.

RyanMG:
Overall: I left notes on the README updates after discussing with @jskupsik . My opinion overall through is not to bikeshed the README for this PR. I think it merits a bigger think about how we we want to present toolbox to somone visting the repo for the first time, and what we would want their "onboarding" experience to me. And this task is more something we would want to ticket out as tasks to look at as a company over time - adding documentation pages, wikis, online introduction videos, etc.

jskupsik:
Updated the build on teamcity: image

PR #752, opened 2025-03-13
Contact app mobile

RyanMG:
As an onboarding exercise, it was suggested that I would build a mobile version of the existing toolbox contacts app. I used this to
1. Get familiar with how toolbox instantiates entrypoints / app routes
2. Learn how to build UI / model data / bindings / etc within the hoist FE exosystem
3. Get familar with as many interfaces as I could (typing, styling, property options, etc)

I pulled as much of the common parts of the desktop contact app as I could - specifically sharing SCSS definitions where overrides or replacements were not needed, and sharing the contact service, which as an API should behave the same between the versions.

However, sharing implementation code proved messier, and may. not be possible in any way where its complexity would be justified in the saved footprint. Currently if I try to make this "common" component builder step, the code fails to execute with an error that a mobile application is importing desktop components (and vice versa). This will mean that for now, any code updates may need to be done in two separate places, violating a pretty basic DRY principle.

I will provide inline comments to this code where specificity is merited, but a few top level notes here:

1. Any new application entry point into toolbox appears to need a new entry to the OAuth provider (Auth0 in this case?) for a valid redirect URL. I am getting a 403 currently when the application refreshes and requests an updated accessToken (`https://login.xh.io/oauth/token`), and this appears to stem from it not liking my provided `redirect_uri` (in this case `http://localhost:3000/mobileContacts/`). I worked around this for my work by just disabling outh in my local hoist-core, but we will need to look into this if any new routes are merged
2. Would like to pick someone's brain on the webpack dev server. I hit an issue at one point where I accidentally entered `/contactsMobile` (plural contactS) rather than `/contactMobile`, and webpack was serving me the chunkhash for the standard contact page. It seems the webpack server does some sort of pattern matcher for entry points where it finds the closest entry point matching the provided pathname left-to-right. So for instance, I can load `/app` as I would expect, but `/apppppppp` would also load the base app entry point. `/ap` however would not, since it does not make an exact match before the pathname ends. _Not critical, just weird and I could not find the cause. Would be good to better understand._
3. I'll need to do more reviewing / get some pointers on how persistence is set up. You will see that I ended up using localstorge to persist user favorites. I was not seeing that working in the existing contacts app (on dev, not just local). It works properly in both places with this change, but I am not doing it in what appears to be the prescribed way. It also appears that this persistance is something I will be using all the time here, so the sooner I get a handle on it the better.

### Note:
The inline comments on this PR are almost all still valid, despite being noted as "outdated". Lee suggested reconfguring my folder choices to create a root "contact" folder which can house common / shared files, then within create "desktop" and "mobile" folders for respective, dedicated code. This was a full refactor of file structure, which broke the previous comments - hense them showing as outdated.


client-app/src/examples/contact/svc/ContactService.ts
```diff
@@ -52,6 +53,11 @@ export class ContactService extends HoistService {
         const {userFaves} = this,
             isFavorite = userFaves.includes(id);
 
+        XH.setPref(PERSIST_APP.prefKey, {
```
RyanMG, line 56: As noted above, I will need some help with persistence management.

Existing (desktop) contacts is saving grid / tile view, but is NOT saving favorites. I see the userFaves being wired up, but then the persistence code never gets called subsequently when the state changes on an item. I am unsure why.

So this resolves it - call persistence manually to save the value. But ultimately I should understand why its not working by default and resolve it that way.

client-app/src/examples/contactMobile/ContactPage.scss
```diff
@@ -0,0 +1 @@
+@import '../contact/DirectoryPanel';
```
RyanMG, line 1: No change to CSS, but I do need it present / loaded.

client-app/src/examples/contactMobile/ContactsPageModel.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    clearCurrentSelection() {
+        this.gridModel.clearSelection();
+    }
+
+    @action
+    setDisplayMode(value: 'grid' | 'tiles') {
+        this.displayMode = value;
+
+        XH.setPref(PERSIST_APP.prefKey, {
```
RyanMG, line 56: See https://github.com/xh/toolbox/pull/752/files#r1994240772. This was persisting correctly alone, but if I include the above fix to user favorites and NOT this, then saving a favorite will clobber the save of mode.

client-app/src/examples/contactMobile/cmp/TileView.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    return vbox({
+        style: {backgroundImage: `url(${profilePicture})`},
+        className: `tb-contact-tile ${isSelected ? 'tb-contact-tile--selected' : ''}`,
+        items: [
+            div({
+                className: 'tb-contact-tile__bar',
+                // This inline style defintion keeps me from needing to fight with
+                // existing styles when making the mobile button want to center itself
+                style: {alignItems: 'center'},
```
RyanMG, line 34: I had a fair bit of fighting the styling while working on this.

We define some styles in SCSS files to customer a component or layout. Said component / layout will already have some styling present from existing styles from XH / TB / BP. Then additionally you can define some styles on an element component which are writing direct to the styles tag. 
This ultimately meant for me that any styles I applied in an SCSS file were clobbered right off by either an existing style tag or - more often - existing styles with far greater specificity. In this case, using an inline style was of highest import, and so was allowing this align-items style to win over the native styles applied to a mobile button.

client-app/src/examples/contactMobile/details/ContactDetails.scss
```diff
@@ -0,0 +1,48 @@
+.tb-mobile-contact-details-panel {
```
RyanMG, line 1: This file is not entirely different from the desktop styles, but for mobile I needed CSS animations. Trying to inhert the common parts was proving troublesome due to the nesting.

client-app/src/examples/contactMobile/details/ContactDetails.ts
```diff
@@ -0,0 +1,145 @@
...
+            position: 'absolute',
+            title: currentRecord?.data.name,
+            headerItems: [button({icon: Icon.close(), onClick: () => model.clearCurrentRecord()})],
+            item: currentRecord ? contactProfile() : null,
+            bbar: bbar()
+        });
+    }
+});
+
+const contactProfile = hoistCmp.factory({
```
RyanMG, line 32: All the definitions below here are another instance where they LOOK the same as the desktop counterparts, but use MOBILE versions of components.

client-app/src/examples/contactMobile/details/ContactDetailsModel.ts
```diff
@@ -0,0 +1,87 @@
...
+        this.formModel.readonly = true;
+        this.formModel.init(this.currentRecord.data);
+    }
+
+    async toggleEditAsync() {
+        const {formModel, contactPageModel, currentRecord} = this;
+        const {readonly, isDirty} = formModel;
+
+        // Pulled from standard details page
+        // How could the form be "dirty" if it's readonly?
```
RyanMG, line 74: Unsure if I am missing something in the isDirty logic?

client-app/src/examples/contact/mobile/DirectoryPanelModel.ts
```diff
@@ -0,0 +1,120 @@
+import {HoistModel, managed, persist, XH} from '@xh/hoist/core';
+import {bindable, makeObservable, observable, action, runInAction} from '@xh/hoist/mobx';
+import {GridModel} from '@xh/hoist/cmp/grid';
+import {nameCol, locationCol} from '../../../core/columns';
+import {uniq} from 'lodash';
+
+// Shared from the desktop version
+import {PERSIST_APP} from '../AppModel';
+
+import ContactDetailsModel from './details/DetailsPanelModel';
```
TomTirapani, line 10: Importing this under a different name to the class is confusing - let's use the same name in both the definition and its usages.

client-app/src/examples/contact/mobile/details/DetailsPanel.ts
```diff
@@ -0,0 +1,145 @@
...
+import DetailsPanelModel from './DetailsPanelModel';
+import './DetailsPanelMobile.scss';
+
+export const detailsPanel = hoistCmp.factory({
+    model: uses(DetailsPanelModel),
+    render({model}) {
+        const {currentRecord, visible} = model;
+
+        return panel({
+            className: `tb-mobile-contact-details-panel${visible ? '--details-visible' : ''}`,
```
TomTirapani, line 21: We should use a mobile navigator to handle drilldowns. This will allow us to:
+ Make the details its own page, meaning you don't need to handle visibility / transitions
+ Route to specific contacts by name
+ Support gesture navigation on device, etc

client-app/src/examples/contact/mobile/AppComponent.ts
```diff
@@ -0,0 +1,22 @@
+import {hoistCmp, uses} from '@xh/hoist/core';
+import {appBar} from '@xh/hoist/mobile/cmp/header';
+import {panel} from '@xh/hoist/mobile/cmp/panel';
+import {Icon} from '@xh/hoist/icon';
+import {AppModel} from '../AppModel';
+import {contactsPage} from './DirectoryPanel';
```
TomTirapani, line 6: I've said elsewhere, but renaming on import is confusing - lets keep import names, file names and class names consistent throughout

TomTirapani (changes requested):

client-app/src/examples/contact/mobile/DirectoryPanelModel.ts
RyanMG: This must have been a missed update. Originally the files had bespoke names in the mobile app, but in a second pass I unified them with the desktop naming. Must have missed this change to the name reference. Ill update.

client-app/src/examples/contact/desktop/AppComponent.ts
```diff
@@ -14,6 +15,13 @@ export const AppComponent = hoistCmp({
         return panel({
             tbar: appBar({
                 icon: Icon.contact({size: '2x', prefix: 'fal'}),
+                leftItems: [
+                    a({
```
RyanMG, line 19: New link to let users access mobile contacts from the existing toolbox UI

client-app/src/examples/contact/mobile/AppModel.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    override async doLoadAsync() {
+        try {
+            this.contacts = await XH.contactService.getContactsAsync();
+
+            runInAction(() => {
+                this.tagList = uniq(this.contacts.flatMap(it => it.tags ?? [])).sort() as string[];
+            });
+            this.directoryPanelModel.loadAsync();
```
RyanMG, line 88: App fetches contacts, then tells the directory to load its grid.

client-app/src/apps/contact.ts
```diff
@@ -2,8 +2,8 @@ import '../Bootstrap';
 
 import {XH} from '@xh/hoist/core';
 import {AppContainer} from '@xh/hoist/desktop/appcontainer';
-import {AppComponent} from '../examples/contact/AppComponent';
-import {AppModel} from '../examples/contact/AppModel';
+import {AppComponent} from '../examples/contact/desktop/AppComponent';
```
RyanMG, line 5: You will se this update across this PR. 
After some back and forth, Lee suggested that I nest the new `mobile` contacts under the parent /examples/contacts, and do the same with `desktop`. 
This means that some of this PR will be just file path adjustments to accomodate this change.

client-app/src/examples/contact/desktop/AppComponent.ts
```diff
@@ -2,18 +2,27 @@ import {hoistCmp, uses} from '@xh/hoist/core';
...
 import {directoryPanel} from './DirectoryPanel';
-import '../../core/Toolbox.scss';
+import '../../../core/Toolbox.scss';
 
 export const AppComponent = hoistCmp({
     displayName: 'App',
     model: uses(AppModel),
 
     render() {
+        const domain = window.location.origin;
```
RyanMG, line 15: Unsure if there is a 'hoist' way to access the URI.

client-app/src/examples/contact/desktop/AppModel.ts
```diff
@@ -1,8 +1,6 @@
 import {XH} from '@xh/hoist/core';
-import {ContactService} from './svc/ContactService';
-import {BaseAppModel} from '../../BaseAppModel';
-
-export const PERSIST_APP = {prefKey: 'contactAppState'};
```
RyanMG, line 5: As desktop AppModel is no longer the root for all of contacts, I moved this to [here](https://github.com/xh/toolbox/pull/752/files#diff-59d924fa7e5d7a8e651c4f0064a0dd29045a0743817cc572ce38000b1d9fe89bR5), which is common to both versions.

client-app/src/examples/contact/mobile/AppModel.ts
```diff
@@ -0,0 +1,95 @@
...
+        makeObservable(this);
+        this.directoryPanelModel = new DirectoryPanelModel(this);
+        this.detailsPanelModel = new DetailsPanelModel(this);
+    }
+
+    async updateContactAsync(id: string, data: PlainObject = {}) {
+        // Mobile select only allows single selections, while desktop allows multiple
+        // This means we receive an array on the desktop picker, and a single value here.
+        // Convert single result into an array of one to keep data consistent.
+        if (data.tags) {
```
RyanMG, line 62: Comment above explains this. At some point maybe mobile needs to be enhanced to support multi-select, but that seems out of scope here.

gradle.properties
```diff
@@ -3,7 +3,7 @@ xhAppName=Toolbox
 xhAppPackage=io.xh.toolbox
 xhAppVersion=7.0-SNAPSHOT
 
-hoistCoreVersion=29.0-SNAPSHOT
+hoistCoreVersion=30.0-SNAPSHOT
```
RyanMG, line 6: I didnt make this upate. This is what is on `develop`
https://github.com/xh/toolbox/blob/develop/gradle.properties#L6
When I merged up this appeared. I rebased origin/develop into local/develop, then merged local/develop into this branch. That has never before made a commit list which was not mine.

client-app/src/examples/contact/mobile/AppModel.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    constructor() {
+        super();
+        makeObservable(this);
+        this.directoryPanelModel = new DirectoryPanelModel(this);
+        this.detailsPanelModel = new DetailsPanelModel(this);
+    }
+
+    async updateContactAsync(id: string, data: PlainObject = {}) {
```
TomTirapani, line 58: Is `data` optional here? I don't see why you call this without `data`?

If not, I'd recommend we lose the default `{}` value, and instead return early if `data` is null, to avoid the call for to the server for a no-op.

RyanMG: Done.

lbwexler:
This training project got superceded by new mobile toolbox --- next standalone app will be portfolio viewer!

PR #753, opened 2025-03-13
Stop news app items from overflowing their parent container

RyanMG:
The news page loads visually correct initially, but on scrolling items out then back onto screen we would see the content of an individual item overflow the parent.

**_Initial:_**
Screenshot 2025-03-13 at 3 00 12 PM

**_After scrolling out then back into view:_**
Screenshot 2025-03-13 at 3 00 19 PM

This is a side effect of the virtual pagination used by AG grid. Technically these items were already breaking out of their parent's container on load. We just dont see it because the items were stacked top to bottom in their containing DIV, and so each item's z-index was lower than the next item in the list, and said next item would cover the accidental overflow.

Once we start to scroll, AG grid starts shifting / pushing grid element into and out of the parent, and using translation CSS definitions to keep them visually positioned correct in the overall list. However, this does NOT manage relative z-indexing. So the top item in the list may become the last item in the element stack, and while its X/Y position is correct, its relative Z position no longer is.

The quickest fix here is to clip all overflow for each item on the parent itself.

### DIFF
--- a/client-app/src/examples/news/NewsPanelItem.scss
+++ b/client-app/src/examples/news/NewsPanelItem.scss
@@ -1,4 +1,8 @@
 .tb-news-panel {
+  & .ag-row {
+    overflow: hidden;
+  }
+
   .news-item {
     p,
     span {


TomTirapani:
Looks great and solves the issue! Thanks for addressing 👍

PR #756, opened 2025-04-01
Suggested adjustments to mobile contact app

TomTirapani:
+ Use AppModel.contacts as the sole observable source of truth
+ Grid records observe changes to AppModel.contacts
+ Detail panel observes changes to AppModel.contacts

### DIFF
--- a/client-app/src/examples/contact/mobile/AppModel.ts
+++ b/client-app/src/examples/contact/mobile/AppModel.ts
@@ -6,7 +6,7 @@ import directoryPanel from './DirectoryPanel';
 import detailsPanel from './details/DetailsPanel';
 import DetailsPanelModel from './details/DetailsPanelModel';
 import DirectoryPanelModel from './DirectoryPanelModel';
-import {observable, runInAction} from '@xh/hoist/mobx';
+import {makeObservable, observable, runInAction} from '@xh/hoist/mobx';
 import {uniq} from 'lodash';
 
 import {
@@ -50,17 +50,14 @@ export default class AppModel extends BaseAppModel {
 
     constructor() {
         super();
+        makeObservable(this);
         this.directoryPanelModel = new DirectoryPanelModel(this);
         this.detailsPanelModel = new DetailsPanelModel(this);
     }
 
-    setCurrentRecord(contact: PlainObject) {
-        this.detailsPanelModel.setCurrentRecord(contact);
-    }
-
     async updateContactAsync(id: string, data: PlainObject) {
         await XH.contactService.updateContactAsync(id, data);
-        this.directoryPanelModel.updateContact(id, data);
+        await this.refreshAsync();
     }
 
     async toggleFavorite(id: string) {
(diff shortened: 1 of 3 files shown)


client-app/src/examples/contact/mobile/AppModel.ts
```diff
@@ -50,17 +50,14 @@ export default class AppModel extends BaseAppModel {
 
     constructor() {
         super();
+        makeObservable(this);
```
TomTirapani, line 53: We need to add `makeObservable` here for the `tagList` and `contacts` observables above

client-app/src/examples/contact/mobile/AppModel.ts
```diff
@@ -75,17 +72,16 @@ export default class AppModel extends BaseAppModel {
...
     }
 
     override async doLoadAsync() {
         try {
-            this.contacts = await XH.contactService.getContactsAsync();
-
+            const contacts = await XH.contactService.getContactsAsync();
             runInAction(() => {
-                this.tagList = uniq(this.contacts.flatMap(it => it.tags ?? [])).sort() as string[];
+                this.contacts = contacts;
```
TomTirapani, line 82: We need to write observables within an action. Alternatively, we could declare them as @bindable

client-app/src/examples/contact/mobile/AppModel.ts
```diff
@@ -75,17 +72,16 @@ export default class AppModel extends BaseAppModel {
...
+                this.tagList = uniq(contacts.flatMap(it => it.tags ?? [])).sort() as string[];
             });
-            this.directoryPanelModel.loadAsync();
```
TomTirapani, line 88: Since we have an observable list of the all the data, we can react to that to load data into the grid rather than loading it directly.

client-app/src/examples/contact/mobile/DirectoryPanelModel.ts
```diff
@@ -72,10 +72,6 @@ export default class ContactsPageModel extends HoistModel {
         });
     }
 
-    updateContact(id: string, data: PlainObject) {
```
TomTirapani, line 75: Removing this in favour of refeshing AppModel.contacts after each change and then loading into the grid using the reaction above. This ensures the state is always consistent throughout the app.

client-app/src/examples/contact/mobile/details/DetailsPanelModel.ts
```diff
@@ -21,6 +18,11 @@ export default class DetailsPanelModel extends HoistModel {
         return !this.formModel.readonly;
     }
 
+    get currentContact() {
+        if (XH.routerState?.name !== 'default.details') return null;
+        return this.appModel.contacts.find(it => it.id === XH.routerState.params.id);
```
TomTirapani, line 23: Rather than giving this its own observable state, we can pull the currentContact straight out of AppModel.contacts. This helps ensure the contact info is always synchronised throughout the app.

client-app/src/examples/contact/mobile/details/DetailsPanelModel.ts
```diff
@@ -39,20 +41,16 @@ export default class DetailsPanelModel extends HoistModel {
...
-    }
-
-    @action
-    setCurrentRecord(contact: PlainObject) {
-        this.currentContact = contact;
-        if (!isNil(contact)) {
-            this.formModel.init(contact);
-        }
+        this.addReaction({
+            track: () => this.currentContact,
```
TomTirapani, line 46: Similarly with initializing the form - we react to an observable getter that is itself derived from AppModel.contacts

TomTirapani (commented):
These changes are mostly concerned with consolidating where the app keeps its state.

Previously, AppModel.contacts, the DirectoryPanel's grid and the DetailsPanel's currentContact were three different instances of contact state that had to be manually kept in sync.

In general it is safer and cleaner to keep such state in one place that serves as the "source of truth", and then derive other state from that master list.

PR #757, opened 2025-04-04
Skip cache testing service inits in local dev by default

amcclain:
- 15s wait introduced by TestCacheService seems unfortunate for every local start of toolbox - disable both of these services by default, but provide instance config to allow a dev who wants to run them locally to keep them on.


amcclain:
I had noticed a little while ago that Toolbox seemed to be halting on each startup - finally looked and was clear why. 

I'm not too clear on the intended use of these services - is this something we wanted to be running always locally?  If they failed for some reason (locally or on deployed instance), how would we know?  Could we add a doc comment to each service to explain?

RyanMG (approved):

PR #758, opened 2025-04-07
Support EntraID as alt OAuth provider

amcclain:


grails-app/services/io/xh/toolbox/security/AuthenticationService.groovy
```diff
@@ -85,7 +112,7 @@ class AuthenticationService extends BaseAuthenticationService  {
         return user?.checkPassword(password) ? user : null
     }
 
-    private static boolean getUseOAuth() {
+    private boolean getUseOAuth() {
```
amcclain, line 115: fold into `getInstanceConfig('oauthProvider') == 'NONE'`

amcclain:
@lbwexler this is ready to go from my point-of-view - a first take on it, as discussed.  Could look in the future at making this easier to switch around for deployed instances via soft-config, or support both at once somehow.

PR #766, opened 2025-05-30
Incorporate settleTime breaking change

ghsolomon:


ghsolomon:
See breaking change introduced by https://github.com/xh/hoist-react/pull/4008

PR #806, opened 2026-02-07
Add `card` and `formFieldSet` examples

amcclain:


client-app/src/desktop/tabs/layout/widgets/ChartWidget.ts
```diff
@@ -25,15 +25,17 @@ export const chartWidget = hoistCmp.factory({
             model: panelModel,
             ...(panelModel.isModal ? modalOpts : {}),
             item: chart(),
-            bbar: [
```
amcclain: Not sure what these changes are - a fix we should take here, or port to another change? Did not want to lose it, but not part of the intended PR

client-app/src/desktop/tabs/other/LeftRightChooserPanel.tsx
```diff
@@ -59,7 +59,7 @@ export const leftRightChooserPanel = hoistCmp.factory({
                 }),
                 bbar: [
                     leftRightChooserFilter({
-                        anyMatch: model.anyMatch
+                        matchMode: model.anyMatch ? 'any' : 'startWord'
```
amcclain: this is unrelated but does look like a good update

PR #811, opened 2026-02-17
Hoist apps should overwrite the app vars, not the vars prefixed with `--xh-`

jskupsik:
This toolbox code was following an old pattern / anti-pattern where the hoist app was directly overwriting the `--xh-` style vars. Hoist react styles/vars makes sure to look for variants without the `--xh-` prefix specifically for this purpose, we should use those instead.


lbwexler:
@jskupsik -- this looks like a super simple one -- do you still want to merge this?

PR #839, opened 2026-04-15
Upgrade build toolchain to JDK 25

lbwexler:
## Summary

- Upgrades the build toolchain (and CI) from JDK 17 to **JDK 25**. Toolbox is an app, not a library, so it compiles and runs against the full JDK 25 API.
- Upgrades Grails `7.0.7 → 7.1.0`.
- `CLAUDE.md` notes the new JDK 25 requirement for local dev.

## Test plan

- [x] `./gradlew clean build` green on JDK 25 against the paired hoist-core branch (composite build)
- [x] `./gradlew bootRun` boots clean on JDK 25, all services init without errors
- [x] `--add-opens` set in `build.gradle` still works — no reflective-access warnings under JDK 25
- [x] Manual UAT: `/app`, `/admin`, `/mobile`, example apps, and Hazelcast cluster admin console on JDK 25
- [x] CI bootRun / snapshot build goes green

## Follow-ups not in this PR

- Virtual threads (`spring.threads.virtual.enabled: true`) — single-property enable, biggest potential win for I/O-heavy workloads, but Hibernate 5.6 has `synchronized` blocks that can pin virtual threads on JDK 25. Wants its own load-test evidence PR.
- Generational ZGC — one JVM arg, marginal for toolbox, save for latency-sensitive apps.
- `--add-opens` audit to prune flags that may no longer be required under JDK 25's module layout. Deserves a focused test pass (cluster rebalance, JMX, metrics endpoints, Netty direct-buffer paths).

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,12 +3,12 @@ xhAppName=Toolbox
 xhAppPackage=io.xh.toolbox
 xhAppVersion=9.0-SNAPSHOT
 
-grailsVersion=7.0.7
-hoistCoreVersion=38.0-SNAPSHOT
+grailsVersion=7.0.8
+hoistCoreVersion=39.0-SNAPSHOT
 dotEnvGradlePluginVersion=4.0.0
 hazelcast.version=5.6.0
 
-runHoistInline=false
+runHoistInline=true
 enableHotSwap=false
 localDevXmx=2G
 
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@ xhAppName=Toolbox
 xhAppPackage=io.xh.toolbox
 xhAppVersion=9.0-SNAPSHOT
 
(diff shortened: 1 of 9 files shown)


gradle.properties
```diff
@@ -3,12 +3,12 @@ xhAppName=Toolbox
...
 
-grailsVersion=7.0.7
-hoistCoreVersion=38.0-SNAPSHOT
+grailsVersion=7.0.8
+hoistCoreVersion=39.0-SNAPSHOT
 dotEnvGradlePluginVersion=4.0.0
 hazelcast.version=5.6.0
 
-runHoistInline=false
+runHoistInline=true
```
jskupsik: Reminder to flip back to `runHoistInline=false` before merge

jskupsik (approved):

PR #868, opened 2026-06-18
Use NPM instead of Yarn

cnrudd:
### Since most client apps are using NPM now, this PR proposes moving Toolbox over to NPM, too.

**4 things to do as part of the cutover:**
1. consider also cutting over hoist-react to NPM
2. update toolbox's deploy script to use NPM, not Yarn
3. cut over the @xh/package-template project to NPM (also need to redo its NPM deploy with github actions)
4. document NPM replacement for `yarn link`, and confirm how link works between an NPM package and Yarn package if we do not cut over hoist-react to NPM.


client-app/package.json
```diff
@@ -62,10 +62,10 @@
     "stylelint-config-standard-scss": "17.x",
     "typescript": "~5.9.2"
   },
-  "resolutions": {
+  "overrides": {
     "@types/react": "18.x",
     "@types/react-dom": "18.x",
+    "@xh/hoist": "^87.0.0-SNAPSHOT",
```
cnrudd, line 68: This hoist override is necessary until "@xh/package-template" is re-released with latest hoist as peer dep.

PR #873, opened 2026-06-24
Demo controls for fmtQuantity `lossless` + `null` precision

lbwexler:
Wires up the **Other › Format Numbers** demo panel to exercise recent formatting changes from hoist-react.

### Changes
- Added `lossless`, `useMillions`, and `useBillions` switches, shown only when `fmtQuantity` is the selected formatter.
- Added a `null` option to the precision picker, exercising `fmtNumber`'s new full-precision handling.
- Added two sample values (`7,100,100` and `7,120,000`) that demonstrate the lossless cutoff (`7,100,100` stays full; `7,120,000` collapses cleanly to `7.12m`).

### Dependency
Depends on the `lossless` option added in the corresponding hoist-react PR. Run with inline hoist (or once a snapshot including it is published) to see the new behavior. Typecheck/lint pass against the current published snapshot since the options object is passed dynamically.


lbwexler:
Merged this, was some additional support for `precision: null.`. But Interesting that for the lossless example, claude generated the exact same examples for ATM and for me!?!

PR #879, opened 2026-07-22
Admin grid tester: Store memory flags, configurable record shape, and a repeatable benchmark harness

amcclain:
Companion to xh/hoist-react#4501. Extends the Admin > Tests > Grid tester into a rig for measuring the memory and load-time profile of a large Store, and for producing numbers worth quoting.

Nothing outside `Admin > Tests > Grid` is touched.

### Data shape controls

The tester can now dial in an arbitrary record shape independently of the fields its Store declares, which is what makes the `useFixedDataShape` criterion (*populated* fields per record, not declared) measurable at all.

- **Extra Fields** / **Populate** — declare N extra fields, and have the server either populate them or leave them defined-but-null. The latter is the wide-and-sparse shape where `useFixedDataShape` is a regression.
- **Value Mix** — `mixed`, `categorical`, `unique` or `numeric`. All four populate the same ~11/12 of the extra fields, so switching mixes varies value *character* without moving populated-field count, which would otherwise confound the two effects.
- **Categories** — cardinality of the categorical string pool, 1 to 1,000,000. Generated names are fixed-width, so a cardinality sweep varies pool size without also varying value byte size. Disabled for mixes that generate no categorical values.

Both new params are strict server-side: an unknown mix or out-of-range cardinality fails the request rather than defaulting, so a mistyped param cannot serve one dataset under another's label.

### Store / fetch flags

Five flags are exposed on their own toolbar row: `useFixedDataShape`, `freezeData`, `retainRaw`, `reuseRecords` (all `StoreConfig`) and `internStrings` (a `FetchOptions` config, applied to whichever load path is active). `useRawAsData` is deliberately not among them — it is incompatible with `useFixedDataShape`, and the cube tester already has its own harness for it.

- The illegal `reuseRecords` + `retainRaw: false` pairing is prevented in the UI, cleared by a reaction, and guarded again at Store construction for configs restored from the ViewManager.
- `freezeData` is a control rather than a hardcoded `false`, defaulted to Hoist's own default so measurements reflect what apps actually run.
- **Toggling `useFixedDataShape` or `freezeData` reloads the app**, behind a confirm dialog that reverts on cancel. This is deliberate and load-bearing: V8 decides property storage per *isolate*, so measuring both record-data representations in one session contaminates whichever runs second. Settings persist across the reload, so a configured A/B survives it. The other three flags only change what is loaded or retained and can be flipped in place.

### Benchmark harness

Opened from the Store toolbar row. Pick a scenario and iteration count, run, read the results grid, "Copy as Markdown" to export. Results persist to local storage, so an A/B survives the reload the record-data flags force.

Three scenarios: **cold load**; **reload (re-fetch)**, the only way to observe cross-fetch interning; and **reload (same raw refs)**, the only way `reuseRecords` can hit, as it matches on raw reference identity.

Each iteration settles the heap, records a baseline, runs one load, settles again, and reports heap delta, bytes/record and load time as min/median/max — never a single number. Uses `window.gc` when Chrome is launched with `--js-flags="--expose-gc"`, otherwise falls back to allocation pressure, and **states which mode it used on every row** so a weakly-settled figure is not mistaken for a real one.

Two properties worth calling out, because both exist to stop the harness reporting something false:

- **Baselines are verified, not assumed.** A pristine empty-grid baseline is captured once per run; later iterations wait for the heap to return to within tolerance of it before measuring. Iterations that never come back still run but are recorded **Suspect**, with the residue in MB, a banner, and a `Heap Δ #1` column holding the first iteration's delta — the one measured from the pristine baseline by construction, and the only trustworthy figure when the median is not.
- **Concurrent runs are refused outright.** Two overlapping runs corrupt both, each clearing the grid under the other, and the suspect check structurally cannot detect it — it only compares a run against its own baseline. Guarded in the model rather than relying on a disabled button.

Rows are self-describing: every flag, record count, declared field count, *measured* populated-field count, value mix and cardinality. Rows recorded before the value-mix work render `?` rather than blank in those columns, with a tooltip noting their byte figures do not compare with later runs.

### Reproducibility

Test parameters save as named ViewManager configs, so a given matrix cell can be restored exactly rather than reconstructed from memory.

### Merge blockers

- **`client-app/tsconfig.json` is committed with the local hoist-react `paths` entry ENABLED**, which is normally kept commented out. The tester sets `Store.useFixedDataShape`, which is not in any published `@xh/hoist` snapshot yet. An in-file note explains. **Re-comment before merge**, once building against a release that includes it.
- Merge only after xh/hoist-react#4501 merges and a snapshot publishes.

### Reviewer note

`GridTestModel.reloadForRecordDataChange` POSTs directly to `xhView/updateState` — a Hoist Core **impl** endpoint — to make the ViewManager's fire-and-forget state write durable before reloading the page. It works and is commented, but it reaches past the public client API. The clean fix is upstream: expose that write as an awaitable task on `ViewManagerModel`. Flagging rather than hiding it; happy to file that and drop this if preferred.

---

Hoist P/R Checklist
-------------------

- [x] Caught up with `develop` branch as of last change.
- [ ] Added CHANGELOG entry, or determined not required. (Admin test-panel tooling only — no app-facing change.)
- [x] Reviewed for breaking changes, added `breaking-change` label + CHANGELOG if so. (None — additive, confined to the grid tester.)
- [x] Updated doc comments / prop-types, or determined not required.
- [x] Reviewed and tested on Mobile, or determined not required. (Desktop admin console only.)


amcclain:
Hoist-react PR: xh/hoist-react#4501

amcclain:
Superseded by #884, which carries this work forward on the `store-fixed-data-shape` branch - renamed to match the hoist-react branch.

Reopening fresh rather than continuing here: this PR's description referred to a `RecordDataFactory` kill switch and a codegen'd factory that no longer exist, and the tester has since grown well past the A/B toggle it started as - configurable value distributions and cardinality, all five Store memory flags, and a benchmark harness with verified baselines.

PR #891, opened 2026-08-19
Added select perf test case, test cases to mobile

jskupsik:
Adds a select performance test case for https://github.com/xh/hoist-react/pull/4590 / https://github.com/xh/hoist-react/issues/4589

Adds a testing catalog for introducing mobile app tests, and adds a mobile counterpart to the above select performance test case.

Some mobile app pages are broken right now - https://github.com/xh/hoist-react/pull/4596 - so this is a draft until mobile is fixed and I can test the harness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Technical
 
 * Migrated client-app package management from yarn classic to pnpm, with the version pinned via `packageManager`. Run `corepack enable pnpm` once before your next install.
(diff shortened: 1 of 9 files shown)


jskupsik:
Tested the mobile test page - select on there seems to work as well as the desktop select.

mobile test directory does not have any in-app listing at the moment and is only accessible by directly going to `/mobile/tests/select`, which should be fine.

PR #895, opened 2026-08-31
Cube tester: PctTotalAggregator rewritten as a calculated field (hoist-react #4620)

lbwexler:
Companion to [xh/hoist-react#4636](https://github.com/xh/hoist-react/pull/4636) (eponymous branch) - exercises `CubeFieldSpec.calculatedFn` in the cube test panel, per the validation plan in [hoist-react #4620](https://github.com/xh/hoist-react/issues/4620).

**Run against a local hoist-react checkout on `claude/issue-4620-kyt4eq`.**

## Changes

* `pctCommission` is now a **calculated field**: percent-of-total computed lazily at read time from the row's SUM-aggregated `commission` and a per-update total memoized in `ctx.appData` via `ctx.filteredRecords`. `PctTotalAggregator` deleted.
* No `pctCommission` values are stamped on source orders in the load or update streams - calculated fields require no source data.
* Column marked `editable: false` (calculated fields are read-only), toggle relabeled `Pct Total (calc)?`, and the field drops out of the query in the non-projection A/B mode (Views with calculated fields require `projectionOnly` connected stores).

## What to verify

- [ ] With `Pct Total (calc)?` on and a streaming update rate set, the View's diagnostics stay on the **data-only** update path (the old aggregator forced a full generation per tick).
- [ ] Visible `pctCommission` cells repaint on every tick, including rows whose own commission did not change (the moving-denominator case) - and leaf rows now show a real percent under `Leaves?`.
- [ ] Sorting by `pctCommission` stays correct across ticks; values match the old aggregator's at parent levels.
- [ ] Heap chart (memory flags) comparable or better vs `develop` at high record multipliers - checks the generated dense/parent-row data classes hold V8 fast-properties shapes.
- [ ] Fund filter + query/grouping changes behave with the calculated field included.

---
_Generated by [Claude Code](https://claude.ai/code/session_012GfHUEN4yQ3jNaAUBZH6uT)_


lbwexler:
Superseded by a follow-up PR from branch claude/issue-4637-kyt4eq, which contains this branch's full history plus the StoreConfig.view API updates.

PR #902, opened 2026-09-10
Add MockDirectoryService, replacing the doLoadUsersForDirectoryGroups override

lbwexler:
Companion to xh/hoist-core#602.

Toolbox mocked only directory group membership resolution, via a `doLoadUsersForDirectoryGroups` override. Group display names and group search went straight to the (disabled) real directory service, so the Admin Console Roles tab reported `No enabled directory service in this application` for every assigned group and returned nothing from its group picker.

The mock directory now sits behind the full `DirectoryService` contract instead - membership, display names, and search - in a new `MockDirectoryService`, still driven by the `mockDirectoryGroups` config and still honoring the `sim_error` name. `RoleService` selects it through hoist-core's `getDirectoryService` seam:

```groovy
protected DirectoryService getDirectoryService() {
    def svc = super.getDirectoryService()
    svc.enabled ? svc : mockDirectoryService
}
```

Delegating to `super` first keeps a real directory service in charge whenever one is enabled - in particular the Entra ID testbed path - and picks up the `directoryGroupProvider` config handling for free.

Requires the hoist-core PR: the `doLoadUsersForDirectoryGroups` extension point is gone there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 -->
 ## 11.0.0-SNAPSHOT - unreleased
 
+### Bug Fixes
+
+* Fixed the Admin Console Roles tab reporting "No enabled directory service in this application" for every assigned directory group, and returning no results from its group search. Toolbox's mock directory now backs those lookups via a new `MockDirectoryService`, alongside the group membership resolution it already provided.
+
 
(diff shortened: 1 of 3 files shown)


lbwexler:
CR: Colin

